PR #32, opened 2023-08-06
File naming fix for analyzers

gr812b:
This should fix the issue of the file naming. It now receives the file name and local path from the backend as part of the response headers


t-wing11 (commented):
code not efficient but i will pass it for now

t-wing11 (approved):
Poor instructions from github admin, this time it'll work

PR #34, opened 2023-08-06
Front end code refactor

t-wing11:
Refactored front end code to separate the CreateGraphModal into there respective page files.

### DIFF
--- a/front-end/src/components/AnalyzersAndSeries.js
+++ b/front-end/src/components/AnalyzersAndSeries.js
@@ -1,6 +1,6 @@
 import '../styles/modalStyles.css';
 
-const AnalyzersAndSeries = ({ dimensions, columns, setDisplayPage, fileTransfer, setShowModal, graphType, liveCheck }) => {
+const AnalyzersAndSeries = ({ dimensions, columns, setDisplayPage, setShowModal, handleSubmit }) => {
 
     const columnGenerator = (n) => {
         let arr = [];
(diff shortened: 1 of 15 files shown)


front-end/src/components/FileStorage.js
```diff
@@ -103,6 +119,23 @@ const FileStorage = ({ files, selectedFiles, setSelectedFiles }) => {
...
+            <div className="buttonFlexBox">
+        <button className="submitbutton" onClick={() => {setDisplayPage(1)}}>Back</button>
+        <button className="submitbutton" onClick={() => {
+          // OnClick, it should get the selected files from the file storage component
+          //getSelectedFiles();
+          console.log("selected files")
+          console.log(selectedFiles)
+          getHeaders(selectedFiles)
+          // Instead of this function, go through 
+          if (graphType === "XYColour") {
```
gr812b, line 131: This, along with graphType and setDimensions are outdated, should be removed

front-end/src/components/AnalyzersAndSeries.js
```diff
@@ -0,0 +1,135 @@
...
+            "analyze": {
+                "analysis": analyzers.filter(analyzer => document.getElementById(analyzer.name).checked)[0].code,
+                "analyzerValues": getAnalyzerOptions()
+            }
+
+        })
+        console.log(seriesInfo)
+    }
+
+    const handleSubmit = (e) => {
```
gr812b, line 66: handleFileTransfer should be more fleshed out, so that it only needs to be passed information, and handled submit should no longer be needed. Define in app.js to clarify how information is passed

gr812b (changes requested):
Make these alterations to keep the code base very clean and easy to understand, which is the purpose of the issue

front-end/src/components/FileStorage.js
t-wing11: as the author of this issue i believe that i understand the purpose better then anybody

t-wing11:
As the author of the related issue I believe that I understand the purpose better then anybody, but as project lead I will take your recommendations and adjust the code

gr812b (approved):
Looks good, another issue down

PR #35, opened 2023-09-14
Made Check for columns when submitting

t-wing11:
Inputted a check for the column so that you cannot submit without having one selected

Produces an alert when you attempt to do so.

### DIFF
--- a/front-end/src/components/AnalyzersAndSeries.js
+++ b/front-end/src/components/AnalyzersAndSeries.js
@@ -49,16 +49,21 @@ const AnalyzersAndSeries = ({ dimensions, columns, setDisplayPage, fileTransfer,
 
         var selectColumns = [];
         for (let i = 0; i < dimensions; i++) {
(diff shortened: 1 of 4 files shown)


front-end/src/components/AnalyzersAndSeries.js
```diff
@@ -49,20 +49,47 @@ const AnalyzersAndSeries = ({ dimensions, columns, setDisplayPage, setShowModal,
...
             "analyze": {
                 "analysis": analyzers.filter(analyzer => document.getElementById(analyzer.name).checked)[0].code,
                 "analyzerValues": getAnalyzerOptions()
             }
-
         })
         console.log(seriesInfo)
     }
 
+    const handleSubmit = (e) => {
```
gr812b, line 71: Calling this "handleSubmit" when a handleSubmit is passed in as a prop causes an error

gr812b (changes requested):

gr812b (approved):
Looks good

PR #40, opened 2023-09-22
Fixed various bugs and fixed live data

gr812b:
Fixed weird if statement for analyzer options
Altered getFile such that it returns the data instead of setting a state variable, allowing this to be used for liveData Added a secondary wrapper function to format the data so it could be called from multiple locations Fixed a few legacy calls to chartInformation before the refactor


gr812b:
Im merging this because no one wants to review my code (but I did test it and its flawless ofc)

PR #41, opened 2023-09-27
parse suspension travel data as integer

Anonymus1:
Parsing uses analog readings from the hall effect sensor and the parser has been changed to reflect this.


gr812b (approved):
This is good

PR #56, opened 2023-11-27
Xy colour

gr812b:
Added functionality to use XYColour graphs. This allows different line segments and points to be coloured according to a third axis. Further support on analyzers such as interpolation is required, but works for now (perhaps buggy)


front-end/src/components/AnalyzersAndSeries.js
```diff
@@ -3,22 +3,35 @@ import Help from './Help';
...
+                    <div className="boldText">{i === 0 ? "X-Axis" : "Y Axis"} </div>
+                    <select className={i} key={i} defaultValue={JSON.stringify(columns[i])}>
                         {columns.map(column => (
                             <option value={JSON.stringify(column)} key={column.header + column.filename}>{column.filename} - {column.header}</option>
                         ))}
                     </select>
                 </div>);
         }
-        return arr;
+        let arr2 = [];
```
gr812b, line 21: We're going to need to fix this code, but since we will be redoing this anyways, it can stay for now

front-end/src/components/AnalyzersAndSeries.js
```diff
@@ -57,9 +70,10 @@ const AnalyzersAndSeries = ({ dimensions, columns, setDisplayPage, setShowModal,
     return (
         <div className="analyzersAndSeriesContainer">
             <h3>Select Axis</h3>
-            <div className="columnHeaders">
+         
+                {graphType === "colour" ? (setDimensions(3)) : (setDimensions(2))}
```
gr812b, line 74: We can one day move this to be based off of a json file with each type of graph

front-end/src/components/Chart.js
```diff
@@ -1,18 +1,20 @@
 import '../styles/chart.css'
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import Highcharts from 'highcharts'
 import HighchartsReact from 'highcharts-react-official'
 import Boost from 'highcharts/modules/boost';
-//import Papa from "papaparse";
+import HighchartsColorAxis from "highcharts/modules/coloraxis";
+require('highcharts-multicolor-series')(Highcharts);
```
gr812b, line 7: Im sure this is not the right format but Im not sure what the alternative is to match it

front-end/src/components/Chart.js
```diff
@@ -196,19 +270,24 @@ const Chart = ({ chartInformation }) => {
...
             intervalId = setInterval(() => {
                 getFileFormat();
             }, 2000);
         }
 
         return () => clearInterval(intervalId);
-      }, [chartInformation.live]);
+    }, [chartInformation.live]);
+
+    const getColourValue = (minhue, maxhue, minval, maxval, val) => {
```
gr812b, line 287: This isnt used

front-end/src/styles/analyzersAndSeriesStyles.css
```diff
@@ -4,6 +4,13 @@
   justify-content: space-around;
 }
 
+.columnHeaders2 {
```
gr812b, line 7: Why name is so bad please dont do this, this is also code that will be rewritten but cmon

gr812b (commented):
Needs changes, at least add comments for what will be changed and needs to be redone

PR #63, opened 2023-11-29
Help Page Update

gr812b:


API/src/main/java/backend/API/MasterExceptionHandler.java
```diff
@@ -12,32 +12,33 @@
 // This class is used to handle exceptions thrown by the API.
```
gr812b, line 12: No clue why this file "updated"

gr812b (commented):
Good to go

PR #70, opened 2023-11-30
57 Handle Long File Names

camdnnn:
### Overview
Added additional formatting to prevent long file names from affecting size column and other row heights

### Rundown
- Added 'break-all' to file select so that all characters text-wrap
- Added maxWidth to rows so that size doesn't text wrap

### What is Implemented
- [x] Long file names don't cause other formatting to change


gr812b (approved):
Easy and clean solution

PR #72, opened 2023-11-30
58 Folder Sizes on File Storage

camdnnn:
Added folder size to folders and refactored format function to be more robust and simpler


front-end/src/components/FileStorage.js
```diff
@@ -66,9 +61,12 @@ const CustomFileRenderer = (props) => {
 };
 
 const CustomFolderRenderer = (props) => {
-    const { isOpen, browserProps } = props;
+    const { files, isOpen, browserProps } = props;
```
gr812b, line 64: There is a better way than passing in all files. The package used already passes down all the information you need in props, try printing that to console and check it out. No need to pass additional stuff here on top of isOpen, browserProps

front-end/src/components/FileStorage.js
```diff
@@ -3,6 +3,14 @@ import '../styles/fileStorage.css'
 import 'react-keyed-file-browser/dist/react-keyed-file-browser.css';
 import RawFileBrowser, { Icons } from 'react-keyed-file-browser';
 
+const formatSize = (size) => {
```
gr812b, line 6: This is very clean

gr812b (changes requested):

gr812b (approved):
Looks great

PR #74, opened 2023-12-25
60 - Abort Duplicate Series

camdnnn:
Stop duplicate series from being added
Message is only displayed on Add Series, if it done for Next then the alert is ignored

Included some updates to page management

### DIFF
--- a/front-end/src/components/modal/create/AnalyzersAndSeries.js
+++ b/front-end/src/components/modal/create/AnalyzersAndSeries.js
@@ -5,6 +5,15 @@ import '../../../styles/analyzersAndSeriesStyles.css';
 import { useState, useRef, useEffect } from 'react';
 const AnalyzersAndSeries = ({ dimensions, columns, movePage, seriesInfo, setSeriesInfo, setSuccessMessage, setDimensions, graphType }) => {
 
+    // Determines if a series already exists with the same columns and analyzer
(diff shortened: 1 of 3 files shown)


gr812b (changes requested):
Good changes, I like the movePage function, the way the duplicate series popup is handled feels off, should not be an error but rather ignored or using our own popup that goes away, just pass in a colour and set to red, but should not interrupt it.

Also, there is a new error in the console, im not sure which setState is causing it, but here is the error. To reproduce, simply create a graph.

image

```CreateGraphModal.js:29 Warning: Cannot update a component (`App`) while rendering a different component (`CreateGraphModal`). To locate the bad setState() call inside `CreateGraphModal`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    at CreateGraphModal (http://localhost:3000/main.7085d9fcb4b42cffa11e.hot-update.js:416:3)
    at header
    at div
    at App (http://localhost:3000/static/js/bundle.js:45:76)```

gr812b (changes requested):
In the last review "Good changes, I like the movePage function, the way the duplicate series popup is handled feels off, should not be an error but rather ignored or using our own popup that goes away, just pass in a colour and set to red, but should not interrupt it."

The alert is kinda goofy is all, consider swapping?

gr812b (approved):
Good to go!

PR #78, opened 2023-12-27
File browser now uses containing folder as modified date

JAWalmsley:
image

Issue mentions preserving folder structure but changing the date modified is easier and less bulky I would say, could be changed though. Date is taken from the containing folder name


gr812b (changes requested):
Last Modified will most likely never be used, worth changing the column name + variables. Other than that easy and good solution

JAWalmsley:
image
New header title, plus styled the header so it wouldn't change lengths while searching

gr812b (approved):
Looks perfect

PR #80, opened 2024-01-21
77 revamped top bar menu

JAWalmsley:
New top bar using icons instead of text buttons
image
image


gr812b (approved):
Looks fantastic, might one day add a settings button, but this is so so much cleaner than it was before

PR #84, opened 2024-02-22
GPS map view and timing gates

JAWalmsley:


front-end/public/topdown.png
gr812b, line 1: Can we get a version with a white outline? This one seems difficult to view when the car is not moving, some way to differentiate from green, blue, brown, gray backgrounds would be nice

front-end/src/components/App.js
```diff
@@ -49,26 +51,31 @@ const App = () => {
...
+        {modal === 'Create' ? <CreateGraphModal setModal={setModal} setChartInformation={setChartInformation} setSuccessMessage={setSuccessMessage} /> : null}
+        {modal === 'Upload' ? <UploadModal setModal={setModal} setSuccessMessage={setSuccessMessage} /> : null}
         {modal === 'Download' ? <DownloadModal setModal={setModal} /> : null}
         {modal === 'Help' ? <HelpModal setModal={setModal} /> : null}
 
-        <Chart chartInformation={chartInformation} />
+        {window.location.pathname === '/map' ?
+          <MapChart />
+          : <Chart chartInformation={chartInformation} />
+        }
```
gr812b, line 71: I think we should try to use BrowserRouter for page directing, for longterm. See [here](https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/blob/IMU-Testing/front-end/src/components/App.js) for how I implemented it for the IMU.

front-end/src/components/map/MapChart.js
gr812b, line 1: For static styles, I think its best to use stylesheets and classnames to apply them. Keeps the already messy codebase a little cleaner and better for maintainability

gr812b (changes requested):
### Overall review
One of the coolest and user friendly features we have, implemented in an independent way.
#### Features that would be nice to add:
- Play, pause and playback speed
- Colour axis? (User selects specific file (in same bin) to show as colour axis such as brake pres or speed))
#### Bugs:
- Checkpoints can be hit multiple times, I think only first occurence should be registered, others ignored
#### Not a bug but
- Styling is applied in-line in a few places that could be avoided with classNames and style sheets
- This branch will eventually be merged with [this branch](https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/tree/76-multiple-graphs-placement-of-objects), which finally uses a lib folder for helper functions and begins to separate those that just *help*. Would be worth it to create a file in that same directory and remove those functions out to here, with jsdoc descriptions for each. Would simplify and make the react files have react in them, while the general javascript ones be just that. Plan is to make this separation even stronger over time for simplicity of frontend.

front-end/public/topdown.png
JAWalmsley: Added!

front-end/src/components/App.js
JAWalmsley: 👍

JAWalmsley:
Added drawing with ctrl-click for touchpads
Can only visit a checkpoint once per lap
Using react router properly
Added white outline to car icon

front-end/src/components/map/testData.json
gr812b, line 1: Probably worth removing this, if you think its important enough to keep around, throw it into the data folder of the master directory and label

front-end/src/components/map/MapDisplay.js
gr812b, line 1: Thoughts on moving all the constants and functions to a separate file? (in a lib directory)

gr812b (changes requested):
Great fixes, the outlined car is fantastic and its just a little bit quicker to use, plus some of the bug fixes are great. I think fixing up the styling would be nice and then maybe make that extra lib file. You can probably just make the folder and use a new file for it so the merge is easy. 

Then lets get this merged!

JAWalmsley:
Alright, should be all good.
Some styles are left inline because they're actually options to a Leaflet component not a real element so it's hard to target in css

PR #85, opened 2024-03-13
68 - Sync Video to Data

camdnnn:
Synching video to data


front-end/src/components/Chart.js
```diff
@@ -1,16 +1,20 @@
 import '../styles/chart.css'
 import React, { useState, useEffect, useRef } from 'react';
-import Highcharts from 'highcharts'
+import Highcharts, { chart, time } from 'highcharts'
```
gr812b, line 3: What is chart, time?

front-end/src/components/Chart.js
```diff
@@ -31,7 +35,7 @@ const Chart = ({ chartInformation }) => {
             enabled: false
         },
         boost: {
-            enabled: true
+            enabled: false
```
gr812b, line 38: We should add checks to disable this when using video... why did we do this again?

front-end/src/components/Chart.js
```diff
@@ -249,14 +263,104 @@ const Chart = ({ chartInformation }) => {
...
+        const videoStart = new Date(chartInformation.video.fileHeaders[0]).getTime()
+        
+        const tempOffsets = []
+        chartInformation.files.forEach(file => {
+            const fileStart = new Date(file.columns[0].timespan[0]).getTime() // Unix date of first timestamp in file
+            tempOffsets.push(videoStart - fileStart)
+        })
+
+        setOffsets(tempOffsets)
+    }, [chartInformation])
```
gr812b, line 325: Duplicate useEffect on chartInformaton as the one you made above, can we split the logic out into their own functions and call them from the same useEffect?

front-end/src/components/Chart.js
```diff
@@ -249,14 +263,104 @@ const Chart = ({ chartInformation }) => {
...
+
+        return timestamps.indexOf(closestTimestamp);
+    }
+
+    const findPointIndex = (timestampIndex, series) => {
+        const timestampPoint = {x: series.xData[timestampIndex], y: series.yData[timestampIndex]}
+        const point = series.points.find(point => point.x === timestampPoint.x && point.y === timestampPoint.y)
+        return series.points.indexOf(point)
+    }
+
```
gr812b, line 363: Can we separate all the functions above into a separate file and store them in a lib folder? See either Jack's branch or multiple graphs, and make a new file for that folder so merge conflicts dont arise

front-end/src/components/modal/FileStorage.js
```diff
@@ -102,21 +101,21 @@ const CustomHeaderRenderer = (props) => {
...
+    // const [files, setFiles] = useState([])
+
+    // useEffect(() => {
+    //     // Fetch data when the component mounts
+    //     fetch(`http://${window.location.hostname}:8080/files/folder/csv`)
+    //       .then((response) => response.json())
+    //       .then((data) => {
+    //         setFiles(data.files);
+    //       });
+    //   }, []); // Empty dependency array ensures that the fetch is only performed once
```
gr812b, line 118: My understanding here is that we fetch the data when the createGraphModal mounts, so that we don't need to wait for it anymore, right?

front-end/src/components/modal/create/GraphSettings.js
```diff
@@ -27,20 +27,21 @@ const GraphSettings = ({ movePage, setGraphType, setLiveCheck }) => {
         <div className="graphTypesBox">
           <div className="graphOptionsText">Graph Types</div>
           <div className="GraphTypeSelect">
-            <select id="graphTypeSelect" >
+            <select id="graphTypeSelect" value={graphType} onChange={getGraphType}>
```
gr812b, line 30: These graphType changes are not necessary and confusing. Should only need setGraphType, and can use your own getGtaphType function to fetch it in this file, not passed as a prop. Also not needed in this line or elsewhere. Disable the live button makes sense, but you can just set its value back to false rather than passing liveCheck down too

front-end/src/components/modal/create/VideoPlayer.js
```diff
@@ -0,0 +1,90 @@
...
+      setDuration(videoRef.current.duration);
+    }
+
+    videoRef.current.addEventListener('timeupdate', handleTimeUpdate);
+    videoRef.current.addEventListener('durationchange', handleDurationChange);
+    return () => {
+      videoRef.current.removeEventListener('timeupdate', handleTimeUpdate);
+      videoRef.current.removeEventListener('durationchange', handleDurationChange);
+    };
+  }, []);
```
gr812b, line 43: Another duplicate useEffect, which can be done with functions and a single useEffect

front-end/src/components/modal/create/VideoSelect.js
```diff
@@ -0,0 +1,65 @@
+import '../../../styles/modalStyles.css';
+import '../../../styles/videoSelectStyles.css';
+import { AiFillFolder } from 'react-icons/ai';
+
+export const VideoSelect = ({ movePage, selectedVideo, setSelectedVideo, files, filteredFiles, setFilteredFiles, fileTimespans, videoTimespans }) => {
```
gr812b, line 5: Why are filteredFiles here and selectedFiles two separate variables? Can we just update the earlier one, or is there some reason? Should probably name for clarity due to similarity now

gr812b: And then theres selectedVideo, which isnt corresponding to selectedFiles but filteredFiles

front-end/src/index.js
```diff
@@ -2,12 +2,18 @@ import React from 'react';
...
 import App from './components/App';
+import { BrowserRouter as Router, Route, Routes } from 'react-router-dom';
+import VideoPlayer from './components/modal/create/VideoPlayer';
 
 const root = ReactDOM.createRoot(document.getElementById('root'));
 root.render(
   <React.StrictMode>
-    <App />
-    
+    <Router>
```
gr812b, line 11: Dont make changes to index.js. Nothing should be in here. Bring this all to App

gr812b (changes requested):
Lots of my review just require clarification, some changes to make, but the feature itself is beauty. If we can get some of the easy ones done, I can provide clarification on the car ride

API/pom.xml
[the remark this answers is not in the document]
camdnnn: three?

API/src/main/java/backend/API/FileUploadController.java
```diff
@@ -169,29 +264,39 @@ public ResponseEntity<Resource> handleFileRequest(
...
 			// Maybe do the serial stuff here, but definitely look in live folder for data
 		}
+		
+		// For all of the input and output files, add the root location to the front
+		for (int i = 0; i < inputFiles.length; i++) {
+			inputFiles[i] = storageService.getRootLocation().toString() + "/" + storageService.getFileExtension(inputFiles[i]) + "/" + inputFiles[i];
+		}
+		for (int i = 0; i < outputFiles.length; i++) {
+			outputFiles[i] = storageService.getRootLocation().toString() + "/" + storageService.getFileExtension(outputFiles[i]) + "/" + outputFiles[i];
+		}
```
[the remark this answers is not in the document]
camdnnn: I agree

API/src/main/java/backend/API/FileUploadController.java
```diff
@@ -169,29 +264,39 @@ public ResponseEntity<Resource> handleFileRequest(
...
 
 		// Then run the selected analyzer
 		if (analyzer != null && analyzer.length != 0 && analyzer[0] != null) {
 			try {
-			Analyzer.createAnalyzer(analyzer[0], inputFiles, inputColumns, outputFiles,
-					Arrays.copyOfRange(analyzer, 1, analyzer.length)).analyze();
+				Analyzer.createAnalyzer(analyzer[0], inputFiles, inputColumns, outputFiles,
+						(Object[]) Arrays.copyOfRange(analyzer, 1, analyzer.length)).analyze();
```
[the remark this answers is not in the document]
camdnnn: You wrote this, I didn't touch this part

API/src/main/java/backend/API/FileUploadController.java
```diff
@@ -169,29 +264,39 @@ public ResponseEntity<Resource> handleFileRequest(
...
 			// storageService.copyFile(inputFiles[0], outputFiles[outputFiles.length - 1]);
-			outputFiles[outputFiles.length - 1] = "./upload-dir/" + inputFiles[0];
+			outputFiles[outputFiles.length - 1] = inputFiles[0];
 		}
 		// Then return the final file, removing the prefix for upload dir
-		Resource file = storageService.loadAsResource(
-				outputFiles[outputFiles.length - 1].substring(13, outputFiles[outputFiles.length - 1].length()));
+		String filePath = outputFiles[outputFiles.length - 1];
+		Path path = Paths.get(filePath);
+		Path newPath = path.subpath(2, path.getNameCount());
```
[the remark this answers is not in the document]
camdnnn: I agree that we should do that

front-end/src/components/Chart.js
camdnnn: not required, idk why they are there

front-end/src/components/Chart.js
camdnnn: good idea

front-end/src/components/Chart.js
camdnnn: yea we should put these in a utils function

front-end/src/components/modal/FileStorage.js
camdnnn: Yes and this was required so that I can pass in the filtered files since I needed file storage to be able to take in an arbitrary set of files to select from

front-end/src/components/modal/create/GraphSettings.js
camdnnn: I don't understand. It is required in my opinion to handle changing the state of the live check?

front-end/src/components/modal/create/VideoSelect.js
camdnnn: there's no selected files

selectedVideo is required to know which object to highlight
setSelectedVideo is required to set the video that is selected since this is the videoSelect class
files is required since we need to know what files exist to filter them
filteredFiles is required since it is used to decide what folders to show in the available folders column
setFilteredFiles is required since they are calculated when a video is selected

selectedVideo is related to filteredFiles since filtered files are the files that overlap with the selected video

It says files not selected files. This page does not handle selecting of the files, the files is all the files since we need to know what they are for filtering

API/src/main/java/backend/API/FileUploadController.java
```diff
@@ -143,6 +172,72 @@ public ResponseEntity<String> listFileInformation(@PathVariable String filename)
 		return ResponseEntity.ok().headers(responseHeaders).body(fileinfo);
 	}
 
+	// Returns the timespan of all the files in a type folder
+	@GetMapping("/timespan/folder/{foldername:.+}")
+	@ResponseBody
+	public ResponseEntity<fileList> listFolderTimespans(@PathVariable String foldername) throws IOException {
```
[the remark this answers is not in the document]
camdnnn: folder handling definitely needs to be reworked on the backend

API/pom.xml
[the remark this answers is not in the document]
gr812b: Theres three green dependencies?

front-end/src/components/modal/create/VideoSelect.js
gr812b: I meant just renaming

gr812b: change filteredFiles to videoSyncFiles

front-end/src/components/Chart.js
camdnnn: Added video as a chartInformation type and used that to dynamically set boosting

front-end/src/components/Chart.js
camdnnn: lib folder was created

front-end/src/components/Chart.js
camdnnn: duplicate useEffect has been removed

front-end/src/components/modal/create/VideoPlayer.js
camdnnn: removed duplicate useEffect

API/pom.xml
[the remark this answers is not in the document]
camdnnn: You added this dependency
link to the commit -> [Added springboot extension dependency](https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/commit/42ca7d1bbed72f1347dc0950787ecb663115b307)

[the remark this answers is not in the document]
gr812b: Removed

API/src/main/java/backend/API/FileUploadController.java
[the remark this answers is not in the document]
gr812b: Done!

front-end/src/components/modal/create/VideoSelect.js
```diff
@@ -0,0 +1,57 @@
...
+
+export const VideoSelect = ({ movePage, selectedVideo, setSelectedVideo, files, fileTimespans, videoTimespans }) => {
+
+    const [filteredFiles, setFilteredFiles] = useState([]);
+
+    useEffect(() => {
+        setFilteredFiles(filterFiles(selectedVideo, files, fileTimespans))
+    }, [selectedVideo]);
+
+    //render the modal JSX in the portal div.
```
hydrowoxy, line 15: This is entirely insignificant but comment formatting perhaps :>

`// Render the modal JSX in the portal div.`

hydrowoxy (approved):
It looks really sick! I agree with Kai's comments. I didn't find anything else (I did try to look for petty syntactical things but I didn't find much, this is really well done). 

Looks great!!

front-end/src/components/views/VideoPlayer.js
```diff
@@ -0,0 +1,37 @@
...
+        setVideoURL(url);
+
+        return () => {
+          URL.revokeObjectURL(url);
+        };
+      });
+  }, []); // Empty dependency array ensures that the fetch is only performed once
+
+  return (
+    <div className = "background">
```
gr812b, line 25: Feels like there are a lot of extra classes / div here that are unneccesary

front-end/src/index.js
```diff
@@ -5,10 +5,7 @@ import App from './components/App';
 
 const root = ReactDOM.createRoot(document.getElementById('root'));
 root.render(
-  <React.StrictMode>
-    <App />
-    
-  </React.StrictMode>
+  <App />
```
gr812b, line 8: Put this back you bastard

gr812b: This is not what I meant when I said "Fix the linting errors"

front-end/src/lib/chartOptions.js
```diff
@@ -17,7 +17,7 @@ export const defaultChartOptions = {
         enabled: false
     },
     boost: {
-        enabled: true
+        enabled: false
```
gr812b, line 20: Default should be true

boktroioi (approved):
OOO she looking niceeee 😍 

Very very cool nice job!!!

JAWalmsley (commented):
Should probably increase the max file upload in application.properties for uploading videos (they're like 3GB minimum)

JAWalmsley (approved):

JAWalmsley:
Looks sick though this website is actually getting really good

front-end/src/components/modal/create/VideoSelect.js
camdnnn: removed the comment

API/src/main/java/backend/API/storage/StorageService.java
```diff
@@ -30,4 +35,12 @@ public interface StorageService {
 
 	String getLast(String filename);
 
+	LocalDateTime[] getTimespan(String filename);
+
+	LocalDateTime[] getTimespan(String filename, LocalDateTime zeroTime);
+
+	LocalDateTime getZeroTime(Path folder);
+
+	String getFileExtension(String filename);
```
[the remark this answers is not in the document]
camdnnn: renamed to getTypeFolder

front-end/src/index.js
camdnnn: added back in

front-end/src/lib/chartOptions.js
camdnnn: updated to be default true

front-end/src/components/modal/create/AnalyzersAndSeries.js
```diff
@@ -3,7 +3,7 @@ import Help from './Help';
 import analyzerData from '../../analyzerData'
 import '../../../styles/analyzersAndSeriesStyles.css';
 import { useState, useEffect } from 'react';
-const AnalyzersAndSeries = ({ dimensions, columns, movePage, seriesInfo, setSeriesInfo, setSuccessMessage, setDimensions, graphType }) => {
+const AnalyzersAndSeries = ({ dimensions, columns, movePage, seriesInfo, setSeriesInfo, setSuccessMessage, setDimensions, graphType, fileTimespans }) => {
```
t-wing11, line 6: Break the parameters so each is on there own line
ex/ 
const AnalyzersAndSeries = ({
      dimensions,
      columns
      ... etc )} => {

front-end/src/components/modal/create/CreateGraphModal.js
```diff
@@ -6,17 +6,46 @@ import { useRef } from 'react';
...
+import { VideoSelect } from "./VideoSelect";
+import { MAX_VIEWS } from '../../views/viewsConfig';
+import Chart from "../../views/Chart";
+import VideoPlayer from "../../views/VideoPlayer";
+import { replaceViewAtIndex } from "../../../lib/viewUtils";
+import { filterFiles } from "../../../lib/videoUtils";
+import { ApiUtil } from "../../../lib/apiUtils";
 
-export const CreateGraphModal = ({ setModal, setChartInformation, setSuccessMessage, chartInformation, buttonID }) => {
+export const CreateGraphModal = ({ setModal, setViewInformation, setSuccessMessage, viewInformation, buttonID, setNumViews, numViews, selectedVideo, setSelectedVideo }) => {
```
t-wing11, line 17: Break these parameters up as well

front-end/src/components/modal/create/CreateGraphModal.js
```diff
@@ -92,7 +123,8 @@ export const CreateGraphModal = ({ setModal, setChartInformation, setSuccessMess
         }}>Next</button>
       </div>
     </div>,
-    <AnalyzersAndSeries dimensions={dimensions} columns={columns} movePage={movePage} seriesInfo={seriesInfo} setSeriesInfo={setSeriesInfo} setSuccessMessage={setSuccessMessage} setDimensions={setDimensions} graphType={graphType}/>
+
```
t-wing11, line 126: Get rid of this line, we don't break our HTML up so this looks weird

front-end/src/components/modal/create/GraphSettings.js
```diff
@@ -1,14 +1,21 @@
 import '../../../styles/GraphSettingsStyles.css';
 
-const GraphSettings = ({ movePage, setGraphType, setLiveCheck }) => {
-
-  const graphType = () => {
-    setGraphType(document.getElementById("graphTypeSelect").value);
-
+const GraphSettings = ({ movePage, setGraphType, setLiveCheck, selectedVideo }) => {
+
+  const handleTypeSelect = (e) => {
```
t-wing11, line 5: I feel like handleTypeSelect is not the right name for this function, since all your doing is adjusting the livedatacheckbox and not the graph type.

front-end/src/components/modal/create/GraphSettings.js
```diff
@@ -27,20 +34,21 @@ const GraphSettings = ({ movePage, setGraphType, setLiveCheck }) => {
         <div className="graphTypesBox">
           <div className="graphOptionsText">Graph Types</div>
           <div className="GraphTypeSelect">
-            <select id="graphTypeSelect" >
+            <select id="graphTypeSelect" onChange={(e) => { handleTypeSelect(e) }}>
               <option value="line">Line</option>
               <option value="spline">Spline</option>
               <option value="scatter">Scatter</option>
               <option value="coloredline">XYColour</option>
+              <option value="video">Video</option>
```
t-wing11, line 42: I think put video higher on the drop down since things like scatter and spline are never used and kind of irrelevant while video is useful.

front-end/src/components/modal/create/VideoSelect.js
```diff
@@ -0,0 +1,56 @@
...
+        <div className="videoSelectContainer">
+            <h3>Select Video</h3>
+            <div className="videoSelect">
+                <div className="videoContainer">
+                    {videoTimespans.map((videoTimespan, index) => (
+                        <label
+                            key={index}
+                            className={`videoLabel ${selectedVideo === videoTimespan ? 'selected' : ''}`}
+                            htmlFor={`video-${index}`}
+                        >
```
t-wing11, line 25: Don't leave floating div tags like this, close it on the above line

front-end/src/components/views/Chart.js
```diff
@@ -1,33 +1,37 @@
...
-import loadingImg from '../assets/loading.gif';
+import loadingImg from '../../assets/loading.gif';
 // TODO: Fix this import (Why is it different?)
 require('highcharts-multicolor-series')(Highcharts);
 
 HighchartsColorAxis(Highcharts);
 Boost(Highcharts);
 
-const Chart = ({ chartInformation }) => {
+const Chart = ({ chartInformation, video, videoTimestamp, setVideoTimestamp }) => {
```
t-wing11, line 19: setvideotimestamp is never used, why is it being imported here?

front-end/src/components/views/VideoPlayer.js
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useState } from 'react';
+import '../../styles/videoPlayerStyles.css';
+import ReactPlayer from 'react-player';
+import { ApiUtil } from '../../lib/apiUtils';
+
+const VideoPlayer = ({ video, videoTimestamp, setVideoTimestamp }) => {
```
t-wing11, line 6: videotimestamp is not used, why is it being passed?

front-end/src/components/views/VideoPlayer.js
t-wing11: I agree, for one component that is a lot of nested divs

t-wing11 (commented):
Looks very good, I liked the reorganization of the front end codebase with the views. The feature looks very good and good job implementing it into the multiple graphs feature.

A couple thoughts:

- Why are the imports separated in every java file, imports should be together.
- Some general formatting issues I pointed out.
- I would like an explanation of the utils.js files as I don't really get what they do.
- We need to lint our code bad, there are so many conflicting styles right now with spacing, comments and formatting (a problem for a later date).


Overall implementation looks good and I don't have any comments on that.

front-end/src/components/modal/create/GraphSettings.js
gr812b: The name gives context to the event that is passed in, as this is called each time the graph type select is updated. Does that make sense, or do you still think it should be changed?

t-wing11: hmmm, I guess that makes sense. Ngl I think i misunderstood how this function works. I think this is ok now.

gr812b:
@t-wing11 
> * Why are the imports separated in every java file, imports should be together.
If you mean every file should have a centralized list of imports, not how java works, if you're saying remove the spaces in them, generally they're separated between packages, so java.x.x.x will be in a big bunch, then org.spring.x.x.x and finally com.drew.x.x.x etc etc for organization
> * I would like an explanation of the utils.js files as I don't really get what they do.
Lets just sit down some time soon to go over em

front-end/src/components/modal/create/VideoSelect.js
camdnnn: I personally prefer this format since it aligns with how {} are handled

front-end/src/components/views/Chart.js
camdnnn: will be used once chart to video synching is implemented

front-end/src/components/views/VideoPlayer.js
camdnnn: will be used on chart to video synching is implemented

front-end/src/components/modal/create/CreateGraphModal.js
camdnnn: removed line

front-end/src/components/modal/create/GraphSettings.js
camdnnn: moved to 2nd position

front-end/src/lib/chartUtils.js
```diff
@@ -28,8 +26,9 @@ export const getSeriesData = async (response, filename, columns, minMax, chartTy
 
     // If not colour, return values in array to allow for boost
     if (chartType !== "colour") {
+        const timestampOffset = columns[headerIndices.x].header === 'Timestamp (ms)' ? new Date(columns[headerIndices.x].timespan.start + 'Z').getTime() - parseFloat(lines[0][headerIndices.x]): 0
         return lines.map((line) => {
-            return [parseFloat(line[headerIndices.x]), parseFloat(line[headerIndices.y])];
+            return [parseFloat(line[headerIndices.x]) + timestampOffset, parseFloat(line[headerIndices.y])];
```
gr812b, line 31: timestampOffset shouldn't cause us to run through all the data as well, please include this in the parsing directly so it only happens once :)

front-end/src/lib/chartUtils.js
```diff
@@ -54,6 +53,11 @@ export const getSeriesData = async (response, filename, columns, minMax, chartTy
     })
 }
 
+export const getTimestamps = async (text) => {
+    const timestampHeaderIndex = text.trim().split("\n")[0].split(",").indexOf("Timestamp (ms)")
+    return text.trim().split("\n").slice(1).map((line) => parseFloat(line.split(",")[timestampHeaderIndex]))
```
gr812b, line 58: Same thing as above here

gr812b (changes requested):

gr812b (approved):

PR #91, opened 2024-03-19
Create ci.yaml

gr812b:


camdnnn (approved):

camdnnn:
I love lint

front-end/src/assets/icons/download.svg
JAWalmsley, line 1: Is it an issue that the linter is linting our svgs? Might not matter but probably not hard to ignore them either

gr812b: Not sure that it is, I cleaned this up myself because I tried importing it using the same standard as everything else (ECMAScript vs CommonJS / Import vs Require) and with import having different rules, I had to remove some properties.

gr812b: React itself was tossing build errors on this matter

gr812b: Linter only enforced the usage of ECMAScript vs CommonJS

Issue #95, opened 2024-03-28
Allow Manual Timespan Setting for CSV


camdnnn:
Currently it is not possible to sync standalone csv files to a video. This is because there is no associated gps data for the file. This should be updated so that they can manual give the dataviewer a start time for the csv which can then be used to sync the csv to a video

PR #109, opened 2024-04-01
96 - Allow CSV Files

camdnnn:
- [x] Fixed issue where csv files without gps time given were not able to be graphed
- [x] Improved handling of datetime to make sure that all files are timestamp and have a defined timespan

### DIFF
--- a/API/src/main/java/backend/API/FileUploadController.java
+++ b/API/src/main/java/backend/API/FileUploadController.java
@@ -227,18 +227,20 @@ public class FileUploadController {
         paths.forEach(
             path -> {
               Path parent = path.getParent();
-              if (storageService.canComputeTimespan(parent.toString())) {
-                // Updates the parent folder and zero time if the parent folder changes to avoid
-                // recalculating the zero time
-                if (container[0] != parent) {
-                  container[0] = parent;
-                  container[1] = storageService.getZeroTime((Path) container[0]);
-                }
-                // Get the path and filename of each file and print it
-                LocalDateTime[] timespan =
-                    storageService.getTimespan(path.toString(), (LocalDateTime) container[1]);
-                timespans.add(
+              if (parent != null) {
+                if (storageService.canComputeTimespan(parent.toString())) {
+                  // Updates the parent folder and zero time if the parent folder changes to avoid
+                  // recalculating the zero time
+                  if (container[0] != parent) {
+                    container[0] = parent;
+                    container[1] = storageService.getZeroTime((Path) container[0]);
+                  }
+                  // Get the path and filename of each file and print it
+                  LocalDateTime[] timespan =
+                      storageService.getTimespan(path.toString(), (LocalDateTime) container[1]);
+                  timespans.add(
                     new fileTimespan(path.toString().replace("\\", "/"), timespan[0], timespan[1]));
+                }
               }
             });
         break;
--- a/API/src/main/java/backend/API/FileUploadController.java
+++ b/API/src/main/java/backend/API/FileUploadController.java
@@ -238,8 +238,7 @@ public class FileUploadController {
                   // Get the path and filename of each file and print it
                   LocalDateTime[] timespan =
                       storageService.getTimespan(path.toString(), (LocalDateTime) container[1]);
-                  timespans.add(
-                    new fileTimespan(
+                  timespans.add(new fileTimespan(
                       path.toString().replace("\\", "/"), timespan[0], timespan[1]));
                 }
               }
--- a/API/src/main/java/backend/API/FileUploadController.java
+++ b/API/src/main/java/backend/API/FileUploadController.java
@@ -239,7 +239,8 @@ public class FileUploadController {
                   LocalDateTime[] timespan =
                       storageService.getTimespan(path.toString(), (LocalDateTime) container[1]);
                   timespans.add(
-                    new fileTimespan(path.toString().replace("\\", "/"), timespan[0], timespan[1]));
+                    new fileTimespan(
+                      path.toString().replace("\\", "/"), timespan[0], timespan[1]));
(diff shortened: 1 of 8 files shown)


gr812b (changes requested):
I can't seem to get the datetime stuff working on CSVs inside a bin folder with the data there.

My setup > Uploaded a bin and single csv, asked the chart to show me a file from the bin, the time starts at 0 on the x-axis.
Expected -> any graphs without that lone csv should have the shifted datetime?

camdnnn:
> I can't seem to get the datetime stuff working on CSVs inside a bin folder with the data there.
> 
> 
> 
> My setup > Uploaded a bin and single csv, asked the chart to show me a file from the bin, the time starts at 0 on the x-axis.
> 
> Expected -> any graphs without that lone csv should have the shifted datetime?


There is a separate issue for adding the feature of manually synching the CSVs to video. The referenced bug issue is to fix that currently and error occurs when trying to use standalone CSV files. That is fixed and the behaviour you mentioned is the intended behaviour for now since we don't have the ability yet to manually sync it and it has no gps data to auto sync 

Here is the link to the issue for the manual synching: https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/95

camdnnn:
> > I can't seem to get the datetime stuff working on CSVs inside a bin folder with the data there.
> > 
> > 
> > 
> > My setup > Uploaded a bin and single csv, asked the chart to show me a file from the bin, the time starts at 0 on the x-axis.
> > 
> > Expected -> any graphs without that lone csv should have the shifted datetime?
> 
> 
> There is a separate issue for adding the feature of manually synching the CSVs to video. The referenced bug issue is to fix that currently an error occurs when trying to use standalone CSV files. That is fixed and the behaviour you mentioned is the intended behaviour for now since we don't have the ability yet to manually sync it and it has no gps data to auto sync 
> 
> Here is the link to the issue for the manual synching: https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/95

Nvm I misunderstood what you said, I'll look into it

camdnnn:
> I can't seem to get the datetime stuff working on CSVs inside a bin folder with the data there.
> 
> My setup > Uploaded a bin and single csv, asked the chart to show me a file from the bin, the time starts at 0 on the x-axis. Expected -> any graphs without that lone csv should have the shifted datetime?

Fixed now

gr812b:
Fix looks good 👍 
Could we set the graph x-axis to still be datetime, but just without the shifting? I like it represented in hours rather than numbers like 4500k and whatnot. This would be on a graph with timestamp (ms) on x-axis but one or more of the files are not in a folder

camdnnn:
> Fix looks good 👍 Could we set the graph x-axis to still be datetime, but just without the shifting? I like it represented in hours rather than numbers like 4500k and whatnot. This would be on a graph with timestamp (ms) on x-axis but one or more of the files are not in a folder

But the the date is wrong :(
image

gr812b:
Check out both 
- Tooltip.formatter https://api.highcharts.com/highcharts/tooltip.formatter
- xaxis.dateTimeLabelFormatter https://api.highcharts.com/highstock/xAxis.dateTimeLabelFormats
Combining these should let us only show the HH:SS on both as we desire here. Maybe dont need to fiddle with dateTimeLabelFormatter as we never have data wide enough to show date on x-axis, but the tooltip does

boktroioi (approved):

camdnnn:
> Check out both
> 
> * Tooltip.formatter https://api.highcharts.com/highcharts/tooltip.formatter
> * xaxis.dateTimeLabelFormatter https://api.highcharts.com/highstock/xAxis.dateTimeLabelFormats
>   Combining these should let us only show the HH:SS on both as we desire here. Maybe dont need to fiddle with dateTimeLabelFormatter as we never have data wide enough to show date on x-axis, but the tooltip does

Separate datetime formatting has been added for partial datetime charts

gr812b:
Would this work on files in folders but that don't have datetime information in them? This edge case could happen with other people saving data to bin files, then the parser parsing it. (Specifically happened tonight)

camdnnn:
> Would this work on files in folders but that don't have datetime information in them? This edge case could happen with other people saving data to bin files, then the parser parsing it. (Specifically happened tonight)



> Would this work on files in folders but that don't have datetime information in them? This edge case could happen with other people saving data to bin files, then the parser parsing it. (Specifically happened tonight)

camdnnn:
> Member

camdnnn:
> Would this work on files in folders but that don't have datetime information in them? This edge case could happen with other people saving data to bin files, then the parser parsing it. (Specifically happened tonight)

Added a check to the backend that verifies that the timespan can be computed

API/src/main/java/backend/API/FileUploadController.java
```diff
@@ -226,11 +226,12 @@ public ResponseEntity<ArrayList<fileTimespan>> listFolderTimespans(
         Object[] container = {null, null};
         paths.forEach(
             path -> {
-              if (path.getParent() != null) {
+              Path parent = path.getParent();
+              if (storageService.canComputeTimespan(parent.toString())) {
```
gr812b, line 230: If parent is null, error is thrown. Needs to check this before invoking .toString() method. This would activate if you upload a CSV file alone.

gr812b (changes requested):
Doesn't work in expected case, also that change from saying Jan 1 would be nice, will look into

API/src/main/java/backend/API/FileUploadController.java
gr812b: Resolved with 49f9ef799930ef3c8d302104fb1e5249a2915043

gr812b (approved):

PR #111, opened 2024-04-03
108 - Add Timestamp Line

camdnnn:
### Goals Accomplished
- [x] Added a vertical line on the chart matching the timestamp for video sync
- [x] Added a second horizontal one for nontimespan graphs for video sync 

### Minor Improvements:
- [x] Update eslint switch case indenting
- [x] Made custom chart config function for video graphs


front-end/src/components/views/Chart.js
```diff
@@ -130,7 +130,10 @@ const Chart = ({ chartInformation, video, videoTimestamp }) => {
       });
       // Updates the chart to show the point that is closest to the video timestamp
       if (seriesPointIndeces.length === 0) return;
-      chartRef.current.series[seriesPointIndeces[0].series].points[seriesPointIndeces[0].point].onMouseOver();
+      const point = chartRef.current.series[seriesPointIndeces[0].series].points[seriesPointIndeces[0].point];
+      // Update chart options so that point.x is the value of the first plotline
+      setChartOptions(movePlotLine(chartOptions, point.x));
+      point.onMouseOver();
```
gr812b, line 136: I tried commenting out this line and I really like the simplicity of just the bar, what do you think? Imo we should look into displaying the value somewhere else, this could also get you multiple series' values all at once which would be a huge help / epic feature for other purposes too.

front-end/src/lib/chartOptions.js
```diff
@@ -25,6 +25,16 @@ export const defaultChartOptions = {
   }
 };
 
+export const movePlotLine = (chartOptions, x) => {
```
gr812b, line 28: This should be near the bottom of the file, don't need to interrupt the chartOption functionality

gr812b (commented):
I like the chartOptions format, its much cleaner now with better naming. Open to chatting about that tooltip stuff

front-end/src/components/views/Chart.js
camdnnn: removed it

front-end/src/lib/chartOptions.js
camdnnn: moved

JAWalmsley (approved):

front-end/src/components/views/Chart.js
```diff
@@ -130,28 +135,113 @@ const Chart = ({ chartInformation, video, videoTimestamp }) => {
...
+    tempValueLines.push('Timestamp: ' + new Date(newLineX).toUTCString());
+    chartRef.current.series.forEach(series => {
+      const value = values.find(value => value.name === series.name);
+      if (value === undefined) return;
+      tempValueLines.push(`${series.name}: (${value.y})`);
+    });
+    setValueLines(tempValueLines);
+  };
 
+  const nonTimespanUpdate = (videoTimestamp) => {
```
gr812b, line 195: Name

front-end/src/components/views/Chart.js
```diff
@@ -130,28 +135,113 @@ const Chart = ({ chartInformation, video, videoTimestamp }) => {
...
+
+  useEffect(() => {
+    try {
+      setChartOptions(movePlotLines(chartOptions, linePoint.x, linePoint.y));
+    } catch (e) {
+      console.log(e);
+    }
+  }, [linePoint]);
+
+  const timespanUpdate = (videoTimestamp) => {
```
gr812b, line 160: Name

gr812b (changes requested):
Fix up a couple of naming things for clarity, including createGraphModal and such

gr812b (changes requested):
image

Just fix error and good to go 🗡️

gr812b (approved):
Looks good to me

PR #138, opened 2024-11-15
FileUploadResource uses OnStorageError and SneakyThrows

hydrowoxy:
Refactored what was already there with `StorageException` (and some related `IOException`) logic


backend/src/main/java/com/mcmasterbaja/FileUploadResource.java
```diff
@@ -58,17 +61,11 @@ public void uploadFile(
...
         }
         break;
 
       default:
-        try {
-          fileData.close();
-        } catch (IOException e) {
-          throw new StorageException("Failed to close fileData", e);
-        }
+        fileData.close();
```
gr812b, line 68: This is def on me, but I think introducing a method within the storageService to handle this binToCsv stuff might be a better idea. Any business logic should be extracted out of the Resources

backend/package-lock.json
gr812b, line 1: Wut is dis

hydrowoxy: I don't even know...

PR #141, opened 2024-11-17
Try system Java install before going to array

JAWalmsley:
Doesn't overwrite the environment variable at the start, only if startup fails the first time


gr812b (approved):
This is a solid idea, didn't realize this was so easy

PR #152, opened 2024-12-02
analyzers use logger instead of print statements (close #150)

### LINKED ISSUE #150: Use Logger in Analyzers
Currently, we still have cases where `System.out.println()` is being used to write to the console. This is most frequent in all the analyzers. Instead, a logger should be injected and used in the same manner that it is used in the StorageService and the Resource files.

hydrowoxy:
Analyzers have been updated to make use of CDI. Now, each concrete analyzer is an injectable bean, which is created by the improved factory. This lets CDI handle the instantiation and lifecycle of our new beans without issue. This allows the use of injecting other beans within them (namely, the logger), more dynamic use of each analyzer and potential for future improvements (such as an analyzer specific datatype rather than using the sometimes ambiguous `AnalyzerParams`)

### DIFF
--- a/backend/src/main/java/com/mcmasterbaja/analyzer/AnalyzerFactory.java
+++ b/backend/src/main/java/com/mcmasterbaja/analyzer/AnalyzerFactory.java
@@ -1,6 +1,6 @@
 package com.mcmasterbaja.analyzer;
 
-import com.mcmasterbaja.model.AnalyzerEnum;
+import com.mcmasterbaja.model.AnalyzerType;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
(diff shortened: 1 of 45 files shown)


gr812b:
Related to https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/157 in regards to parsing the inputs

backend/src/main/java/com/mcmasterbaja/analyzer/AnalyzerFactory.java
```diff
@@ -1,88 +1,79 @@
...
+  @AnalyzerType(AnalyzerEnum.SGOLAY)
+  Analyzer sgolayAnalyzer;
 
-      case INTERPOLATER_PRO:
-        return new InterpolaterProAnalyzer(inputFiles, inputColumns, outputFiles);
+  @Inject
+  @AnalyzerType(AnalyzerEnum.SPLIT)
+  Analyzer splitAnalyzer;
 
+  public Analyzer createAnalyzer(AnalyzerEnum type) {
```
camdnnn, line 53: I find this confusing since there is now an AnalyzerEnum and a AnalyzerType

backend/src/main/java/com/mcmasterbaja/analyzer/AnalyzerFactory.java
```diff
@@ -1,88 +1,79 @@
...
+        return rdpCompressionAnalyzer;
+      case ROLL_AVG:
+        return rollAvgAnalyzer;
+      case SGOLAY:
+        return sgolayAnalyzer;
+      case SPLIT:
+        return splitAnalyzer;
       default:
-        return null;
+        throw new IllegalArgumentException("Invalid analyzer type");
```
camdnnn, line 76: By extension, I also find this a bit a confusing since there is now and AnalyzerEnum and AnalyzerType

backend/src/main/java/com/mcmasterbaja/analyzer/AnalyzerType.java
camdnnn, line 1: There is probably a good reason for this, I just don't understand why we need a interface of an enum value

backend/src/main/java/com/mcmasterbaja/model/AnalyzerParams.java
camdnnn, line 1: see my previous comments

backend/src/main/java/com/mcmasterbaja/model/AnalyzerEnum.java
camdnnn, line 1: Is there a better way of doing this such that we don't need to define a new enum value for each new Analyzer that we add? I feel like it would be better to avoid having this structure

backend/src/main/java/com/mcmasterbaja/model/AnalyzerEnum.java
```diff
@@ -1,6 +1,6 @@
 package com.mcmasterbaja.model;
 
-public enum AnalyzerType {
+public enum AnalyzerEnum {
```
camdnnn, line 3: I've already said it many times already but I think that we could come up with a better naming or structure for this. I don't like the type being called Enum

camdnnn (changes requested):
Good implementation and code cleanup, I just think that we should take a second look at the analyzer enum structure and naming

backend/src/main/java/com/mcmasterbaja/analyzer/AnalyzerType.java
gr812b: This is actually a qualifier, basically used to annotate beans with a more specific naming when injecting them

backend/src/main/java/com/mcmasterbaja/model/AnalyzerEnum.java
gr812b: Enum is def the best way, we need to have some way to represent the possible options to our clients (any who call the backend). Enum is incredibly standard for this purpose.

gr812b: I believe the alternative would be using strings, such as "sum" or such. I think the biggest downside is not having a strict list of acceptable options here. An enum is much more type safe.

For this PR, were it a string, we'd be able to use the @Named annotation rather than this custom one

camdnnn (approved):
LGTM!

PR #155, opened 2024-12-08
Sass + css modules setup

gr812b:
Migrating to Sass and CSS modules for all our front end styling. Further will create Sass mixins for common properties across the app and handle theming.

Some of it is "imcomplete" in a way, but most of it is just stuff I didn't spend the time on to fix old code. IMO this sets the satndards for SCSS nicely, and we can revisit the css (thats been migrated to scss) in here later on like the graph styling.

Important things are the way the themes are setup (and used in Topbar.scss), the way font sizes are used and generally SASS things like indenting the styles and such.


front-end/src/ThemeContext.tsx
gr812b, line 1: Purpose of the context provider is to allow a shared state of the useTheme hook. This will sync its state no matter which component its used it.

camdnnn (approved):
Looks good to me

PR #156, opened 2024-12-08
Add Desktop Version

camdnnn:


backend/src/main/java/com/mcmasterbaja/binary_csv/BinaryToCSV.java
```diff
@@ -12,13 +12,13 @@ public static native void bytesToCSV(
...
-    if (System.getProperty("os.name").equals("Mac OS X")) {
-      path += relativePath + "/libbinary_to_csv_lib.dylib";
-    } else {
-      path += relativePath + "binary_to_csv_lib.dll";
-    }
+    // Use environment variable if given or default to hardcoded path
+    String path = (resourcePath != null) ? resourcePath : (System.getProperty("user.dir") + relativePath);
+
+    // Use proper file and library based on OS
+    path += System.getProperty("os.name").equals("Mac OS X") ? "/libbinary_to_csv_lib.dylib" : "binary_to_csv_lib.dll";
```
gr812b, line 21: Heads-up I believe you'll have merge conflicts here for the linux case in main

front-end/.gitignore
```diff
@@ -2,6 +2,7 @@
 
 # dependencies
 /node_modules
+/uploads
```
gr812b, line 5: Confused by this, does electron build directly into the front-end dir with the backend? Doesn't really make sense to gitignore backend things in this dir

front-end/package.json
```diff
@@ -24,15 +33,62 @@
...
   "scripts": {
     "start": "vite",
-    "build": "tsc && vite build",
+    "build:frontend": "tsc && vite build",
+    "build:electron": "electron-builder",
     "serve": "vite preview",
-    "lint": "npx eslint . --fix"
+    "lint": "npx eslint . --fix",
+    "electron": "electron src/electron/main.js --trace-warnings",
+    "dev": "concurrently \"npm run start\" \"npm run electron\""
```
gr812b, line 48: Dev is a pretty misleading name here if this is for electron dev stuff

front-end/package.json
```diff
@@ -24,15 +33,62 @@
...
+    },
+    "nsis": {
+      "oneClick": false,
+      "allowElevation": true,
+      "allowToChangeInstallationDirectory": true,
+      "installerIcon": "public/bajacar.ico",
+      "uninstallerIcon": "public/bajacar.ico",
+      "installerHeaderIcon": "public/bajacar.ico",
+      "createDesktopShortcut": true,
+      "createStartMenuShortcut": true,
```
gr812b, line 86: This is fun

front-end/src/components/App.tsx
```diff
@@ -4,13 +4,14 @@ import { UploadModal } from './modal/upload/UploadModal';
...
 import { useEffect, useState } from 'react';
-import { BrowserRouter, Routes, Route } from 'react-router-dom';
+import { BrowserRouter, HashRouter, Routes, Route } from 'react-router-dom';
 import Topbar from './Topbar/Topbar';
 import Views from './views/Views/Views';
 import $ from 'jquery';
 import { MAX_VIEWS } from './views/viewsConfig';
 import Chart from './views/Chart/Chart';
 import MapChart from './map/MapChart/MapChart';
+import { isElectron } from '@lib/apiUtils';
```
gr812b, line 14: All this isElectron stuff needs to be thought out a bit more. The changes to the codebase that change how the dev has to work, just for electron to function is quite annoying + hard to understand why its like that, what changes might break it, etc. If a HashRouter would work for both, we can just use that. If there are downsides, we should rethink our approach

front-end/src/electron/main.js
gr812b, line 1: Should be a ts file

front-end/src/lib/apiUtils.ts
```diff
@@ -1,5 +1,8 @@
 import { AnalyzerType, FileInformation, FileTimespan, MinMax } from '@types';
 
+export const isElectron = typeof process !== 'undefined' && process.versions?.electron;
```
gr812b, line 3: Definitely not the spot for this. This should only be logic to fetch from the Api

package.ps1
```diff
@@ -0,0 +1,50 @@
+# Set global error handling preference
+$ErrorActionPreference = "Stop"
+
+# Define paths
+$backendPath = "backend"
+$frontendPath = "front-end"
+$backendTargetPath = "$backendPath\target"
+$backendDLLPath = "$backendPath\src\main\java\com\mcmasterbaja\binary_csv"
+$frontendBackendPath = "$frontendPath\backend"
```
gr812b, line 9: Bro what is this

package.ps1
gr812b, line 1: Overall this ps1 feels like it has redundant steps. Builds the backend, copies the backend to the frontend dir, lets the frontend build with the backend packaged and then finishes? Why not just build backend, build frontend and package. Also instead of all the copying maybe just passing the path so that the front end doesnt need to gitignore backend files like it has?

gr812b (changes requested):
Overall very great that it works, lots of work went into this. Im a big antifan of the stuff within the frontend to make it work. As a developper, I shouldn't need to test my feature in electron to make sure it works, it should just work. If we need to look into other config stuff or approaches we can do so. Package stuff looks good just some things to clean up

front-end/.gitignore
camdnnn: yes sorry, this was a leftover from testing

front-end/package.json
camdnnn: does anything need to be changed?

front-end/src/components/Topbar/Topbar.tsx
```diff
@@ -41,9 +41,21 @@ const Topbar = ({ setModal, numViews, setNumViews }: TopbarProps) => {
...
+
+  const routePrefix = isElectron ? '#' : '/';
+  const [page, setPage] = useState('');
+  const onIconClick = (path: string) => {
+    if (page != path) {
+      setPage(path);
+      window.location.href = routePrefix + path;
+    } else if (isElectron) {
+      window.location.reload();
+    }
```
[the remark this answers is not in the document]
camdnnn: I think that devs should have to think about how their changes impact the codebase and shouldn't make breaking changes to existing features when possible

package.ps1
camdnnn: this is because the package.json needs all of the files to be in the front-end folder and is not able to access files across to the backend folder making so that they need to be brought over temporarly and this also is needed to make sure that the pathing is correct since the bin is not multiple file layers deep

package.ps1
camdnnn: the paths

front-end/src/components/map/MapDisplay/MapDisplay.jsx
```diff
@@ -134,7 +134,7 @@ const MapDisplay = ({ setLapsCallback, gotoTime }) => {
   // GEOJSON USES LONG, LAT NOT LAT, LONG
 
   const marker = L.icon({
-    iconUrl: '/topdown_outline.png',
+    iconUrl: (isElectron ? __dirname : '') + '/topdown_outline.png',
```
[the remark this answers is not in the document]
camdnnn: I'm fine to update this, I think that the reason that this is the only image that I had to do this for is because it is pathing to it directly instead of using the @assets pathing. @JAWalmsley is there a reason that this path is using the direct pathing instead of using a reference to the object?

backend/src/main/java/com/mcmasterbaja/binary_csv/BinaryToCSV.java
camdnnn: yes merge conflicts were resolved

front-end/.gitignore
camdnnn: removed

package.ps1
camdnnn: updated to use better pathing

package.ps1
camdnnn: updated to use better pathing

front-end/src/components/map/MapDisplay/MapDisplay.jsx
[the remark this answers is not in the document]
camdnnn: Moved the assets to the assets folder from public and then used that so now IsElectron is not being used anymore here

front-end/src/lib/apiUtils.ts
camdnnn: moved into a new electronUtils.ts file

front-end/src/electron/main.js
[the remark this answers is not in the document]
camdnnn: converted to typescript

camdnnn: converted

front-end/package.json
camdnnn: removed as it is not currently used

front-end/package.json
camdnnn: updated to not create start menu shortcut

front-end/src/components/Topbar/Topbar.tsx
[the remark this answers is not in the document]
camdnnn: moved to be handled in utils file

front-end/src/components/App.tsx
camdnnn: moved to be handled in utils file

front-end/package.json
```diff
@@ -24,15 +33,69 @@
...
+    "build:frontend": "tsc && vite build",
+    "build:main": "tsc -p tsconfig.main.json",
+    "build:electron": "electron-builder",
     "serve": "vite preview",
-    "lint": "npx eslint . --fix"
+    "lint": "npx eslint . --fix",
+    "electron": "electron src/main.ts --trace-warnings"
+  },
+  "build": {
+    "appId": "com.dataviewer.mcmaster",
```
gr812b, line 51: Should be com.mcmaster.data-viewer?

front-end/package.json
```diff
@@ -24,15 +33,69 @@
...
+    "nsis": {
+      "oneClick": false,
+      "perMachine": true,
+      "runAfterFinish": false,
+      "allowToChangeInstallationDirectory": true,
+      "createDesktopShortcut": true,
+      "createStartMenuShortcut": false,
+      "installerIcon": "public/bajacar.ico",
+      "installerHeaderIcon": "public/bajacar.ico",
+      "shortcutName": "Dataviewer"
```
gr812b, line 94: Data viewer is 2 words

front-end/tsconfig.main.json
gr812b, line 1: Maybe want to look at this file in terms of whether it applies any changes to typescript across most of the frontend as a whole. Was considering adding this for some of the linting stuff anyways

gr812b (commented):
Everything looks very close to done and very clean. Will just comment since we'll merge after new changes

front-end/package.json
camdnnn: updated

front-end/package.json
camdnnn: updated

front-end/tsconfig.main.json
camdnnn: Yea, idrk. I made it seperate because I needed to perseve the file to make sure that I could path to it

gr812b (approved):

PR #199, opened 2025-01-14
latitude vs longitude 1:1

switzmab:
Ensures longitude and latitude maintain a 1:1 ratio when window is resized


front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -78,7 +78,13 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
   const { width, height, ref } = useResizeDetector({
     onResize: () => {
       if (chartRef.current) {
-        chartRef.current.setSize(width, height);
+        if(width!=undefined && height!=undefined) {
+          const size = Math.min(width ,height);
+          chartRef.current.setSize(size,size);
+        }else {
+          chartRef.current.setSize(width,height);
+        }
```
gr812b, line 86: Faiirrlly certain this would only make the graph itself a square, but not concern itself with the scale of the axes. Taking a look into it now

gr812b: This also still applies to all graphs

gr812b: image

This is just an example, but if you see here the axes not being to scale, then it doesn't exactly solve the issue here. Just making the graph itself a square doesn't resolve the sizes being off regardless. On the y, we have 0-300k and on the x 0-50k, which is a big difference in scale

gr812b (changes requested):

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -2,7 +2,7 @@ import styles from './Chart.module.scss';
 import { defaultChartOptions, getChartConfig, movePlotLineX, movePlotLines } from '@lib/chartOptions';
 import { LIVE_DATA_INTERVAL, validateChartInformation } from '@lib/chartUtils';
 import { useState, useEffect, useRef } from 'react';
-import Highcharts from 'highcharts';
+import Highcharts, { color } from 'highcharts';
```
gr812b, line 5: Why is this imported?

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -74,11 +74,34 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
     return () => clearInterval(intervalId);
   }, [chartInformation, refetch]);
 
-  
   const { width, height, ref } = useResizeDetector({
     onResize: () => {
       if (chartRef.current) {
-        chartRef.current.setSize(width, height);
+        //console.log(chartRef.current.userOptions.xAxis);
+        //console.log(chartRef.current.yAxis);
```
gr812b, line 81: When testing is done, remove console logs, don't just comment out. We only keep console logs if they're in some contentious code or slip through

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -74,11 +74,34 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
...
+        if (width != undefined && height != undefined) {
+          if ((chartRef.current.userOptions.xAxis?.[0].title.text == "GPS LATITUDE" && chartRef.current.userOptions.yAxis?.[0].title.text == "GPS LONGITUDE") || (chartRef.current.userOptions.xAxis?.[0].title.text == "GPS LONGITUDE" && chartRef.current.userOptions.yAxis?.[0].title.text == "GPS LATITUDE")) {
```
gr812b, line 83: Running into issues with the xAxis property existing in the shape of [ {} ] meaning it is a list of a single empty object. This means the .title returns undefined, which then returns an error of .text does not exist on undefined. I believe including a null check (similar to xAxis?) on the title should fix this issue.

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -74,11 +74,34 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
...
+            const aspectRatioB = width / height;
```
gr812b, line 84: Unused?

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -74,11 +74,34 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
...
+            const aspectRatio = chartRef.current.chartWidth / chartRef.current.chartHeight;
```
gr812b, line 85: chartWidth and chartHeight only refer to its actual dimensions, not the scale of the xAxis and yAxis. You'll need to use the limits as I gave you a while back to find the proper aspect ratio.

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -74,11 +74,34 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
...
+            const newWidth = width;
```
gr812b, line 86: This is the same width? What's the point of this extra variable

front-end/src/components/views/Chart/Chart.tsx
```diff
@@ -74,11 +74,34 @@ const Chart = ({ chartInformation, video, videoTimestamp }: ChartProps) => {
...
+            const newHeight = newWidth / aspectRatio;
+            if (Math.abs(chartRef.current.chartWidth - newWidth) > 1 || Math.abs(chartRef.current.chartHeight - height) > 1) {
+              chartRef.current.setSize(newWidth, newHeight);
+            }
+            else {
+              if (Math.abs(chartRef.current.chartWidth - width) > 1 || Math.abs(chartRef.current.chartHeight - height) > 1) {
```
gr812b, line 92: These two conditions for the if statements seem the exact same. If newWidth = width

gr812b: I'm also not certain why you're checking these in the if at all?

gr812b (changes requested):
Seems like the issue is not quite fixed, this solution attempts to maintain the aspect ratio of the chart, but that aspect ratio at the start may be wrong. Its in the right direction of calculating a ratio but that ratio needs to consider the scale of the x axis and y axis. Furthermore, if the screen is made larger, the chart often extends beyond the size of the screen, which isn't ideal. In this case, setting the limits to be larger than originally would allow its shape to be maintained

Issue #168, opened 2025-01-10
UI - Footer


gr812b:
**Feature Description**
Create the footer component from the figma. This includes copyright and title, our logo and the socials. Socials should link to actual mcmaster baja socials using <a> tags. Socials can be found from the [McMaster Baja Racing website](https://mcmasterbaja.ca/).

![image](https://github.com/user-attachments/assets/cd4ad116-aeb5-4540-b6ca-c60da9771eea)

PR #201, opened 2025-01-15
Footer UI update (#168)

hydrowoxy:


front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+
+.footer {
+    @include fonts.nav-title;
+    background: var(--background);
+    display: flex;
+    justify-content: space-between;
+    padding: 1rem 0rem 1rem 0rem;
+    color: var(--text-colour);
+    position: sticky;
+    z-index: 1;
```
gr812b, line 11: Wait for mixin in the modal PR and then use the map from there

front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    .logo {
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        width: 150px;
+
+        img {
+            width: 150px !important;
+        }
+    }
```
gr812b, line 28: I think this is dead css, the .logo className isn't used anywhere. Also use rem instead of px where possible, even for big guys like 150px

front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+        
+        .copyright {
+            @include fonts.text-huge; 
+            color: var(--tertiary);  
+            filter: brightness(2);
+        }
+
+        .title {
+            @include fonts.nav-title;  
+            margin-left: 8px;
```
gr812b, line 43: Use rem (0.5rem in this case works)

front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+
+            &:active {
+                box-shadow: 0px 0px 3px #000000;
+                transition: box-shadow 1s;
+                transition-delay: -0.5s;
+            }
+
+            .icon {
+                width: 1em;
+                height: 1em;
```
gr812b, line 71: Use rem not em

front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+            @include fonts.nav-title;  
+            margin-left: 8px;
+        }
+    }
+
+    .buttons {
+        display: flex;
+
+        > * {
+            padding: 8px 1px;
```
gr812b, line 51: Use rem, also why 1px here?

gr812b: I also think this padding is not needed, just include in the margin instead and vertical padding will be handled by flex box here and align-items

front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+            background: none !important;
+            width: fit-content; 
+            font-size: 2.5rem;
+            margin: 0px 8px !important;
```
gr812b, line 55: Pls no use !important unless absolutely needed, should generally be avoided. Here it accomplishes nothing, also use rem :D

front-end/src/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,78 @@
...
+            align-items: center;
+            display: flex;
+
+            &:hover {
+                filter: brightness(0.8);
+            }
+
+            &:active {
```
gr812b, line 63: I like this

front-end/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <a href="https://x.com/mcmaster_baja" title="Twitter" target="_blank" rel="noopener noreferrer" className={styles.iconLink}>
+          <img className={styles.icon} src={icons['twitter']} alt="Twitter" />
+        </a>
+      </div>
+ 
+    </div>
+  );
+};
+
+export default Footer;
```
gr812b, line 43: The standard we've had so far is to not default export the Footer, might change but for now stick to normal export

gr812b (changes requested):
Mostly just some CSS cleanup, basically all just standards or cleanup. Footer looks great. 

Something you _could_ look into would be the responsiveness, basically that extra padding on the left and right of each child in the main flexbox takes away a lot of space when zooming in. If you put the padding on the outside of that flexbox, then each child doesn't need the padding and it will expand better!

PR #202, opened 2025-01-19
105 subteam geared options on bins

### LINKED ISSUE #105: Subteam Geared Options (On Bins)
Often we find ourselves doing a certain subteams analysis on certain bins. It would be far faster to have the option to run certain bins through select analyzers (on the correct data files) automatically upon upload / at request in order to generate the correct files. A possible other sol;ution would be to add a "user-friendly" createGraphModal variant which brings us through selecting the bin file and which data you want to see (Shift curve, power curve, etc) rather than the individual files, which it can then run automatically on that bin.
- Drivetrain would want Primary RPM (sGolay), Secondary RPM (sGolay), Speed vs Time, Shift Curve (Sec RPM (sGolay) vs Prim RPM (sGolay)), others?
- Suspension would want Suspension Travel (x4) (Cubic analyzer), Strain gauges (?), IMU data (?), GPS Data
- Controls might want brake pressures, IMU data, etc

JAWalmsley:
Closes #105 ,  new flow for creating a graph based on preset analyzer and file lists.
Access with the new button on the topbar in the old UI 
image

image


front-end/src/components/modal/create/PresetSelectionPage/PresetSelectionPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+  selected,
+  clickCallback
+}: {
+    title: string;
+    description: string;
+    selected: boolean;
+    clickCallback: () => void;
+}) => {
+  return (
+    <div className={`presets-preset ${selected ? 'presets-preset--selected' : ''}`} onClick={clickCallback}>
```
gr812b, line 25: Use ClassNames here, slightly cleaner standard going forward! Also should be able to just apply the --selected class and css should apply all the same. 

Also also more of a nitpick, doesn't really matter so feel free to ignore but you can just apply another classname (such as selected directly (no need for --) ) and then target it with &.selected for scss

gr812b: Oh if using css then ignore second part

gr812b: Also remembering this is all just make it work code moreso than final code for new styles so feel free to ignore all this for this PR

front-end/src/types/GraphPreset.ts
```diff
@@ -0,0 +1,18 @@
+import { AnalyzerType } from './ApiTypes';
+
+/**
+ * axes[i] is the name of a data series inside axisFiles[i]
+ */
+export interface GraphPreset {
+    axisFiles: string[],
+    axes: string[],
```
gr812b, line 8: If these have a direct mapping, perhaps we should do this as a list of {axisFile, axis} instead?

front-end/src/types/GraphPreset.ts
```diff
@@ -0,0 +1,18 @@
...
+    analyser: AnalyzerType | null,
+    analyserOptions: string[],
```
gr812b, line 10: Supernit: We use american spelling for analyzer in code so far

front-end/src/types/ChartInformation.ts
gr812b, line 1: Check your autoformatter, indent should be 2 spaces instead of 4, will prevent these random changes :D

front-end/src/lib/videoUtils.ts
```diff
@@ -3,9 +3,13 @@ import { FileInformation, FileTimespan, ChartInformation, ExtSeries } from '@typ
...
 export const computeOffsets = (chartInformation: ChartInformation, videoTimespan: FileTimespan) => {
   const videoStart = new Date(videoTimespan.start).getTime();
-  
+
   const tempOffsets: number[] = [];
   chartInformation.files.forEach(file => {
+    if(file.columns[0].timespan.start == null)
+      {
+        throw new Error('File has no timespan start')
+      }
```
gr812b, line 12: This check is def required, but the case in which it doesn't exist, not sure it should error the application. For me right now its throwing an error on basic files. Maybe just return for right now and we can investigate this more in the future in a separate ticket

gr812b (changes requested):
I lef a few comments still in html and css brained but you can ignore all those comments!

I think the only thing that really needs investigating is that one error case, and then I had a weird bug with the window size of 300 for sGolay. I might need to look into that cause it had some odd behaviour.

Also thank you for catching more error cases in random spots! Always good to have more cleaned up and caught properly

front-end/src/components/modal/create/PresetSelectionPage/PresetSelectionPage.tsx
JAWalmsley: It'll need to be changed for the new UI anyways so I figure I'll leave it for now

front-end/src/types/GraphPreset.ts
JAWalmsley: Good idea, implemented

front-end/src/components/modal/create/SimpleCreateGraphModal/SimpleCreateGraphModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+  };
+
+  const movePage = (amount) => { 
+    setDisplayPage(displayPage + amount);
+  };
+
+  //Stuff for handling final submit
+  const handleSubmit = useCallback(() => {
+    if(selectedPreset == null || selectedBinFile == null)
```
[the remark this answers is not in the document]
JAWalmsley: I do not want to hear code review from you, robot. I am a divine being, you are a mechanism. You have no right to speak in my holy tongue.

gr812b (approved):

PR #203, opened 2025-01-20
Add Text Field

camdnnn:
Added the text field component matching the specification from the Figma design

image


front-end/src/components/App.tsx
```diff
@@ -13,6 +13,8 @@ import MapChart from './map/MapChart/MapChart';
 import cx from 'classnames';
 import ModelViewer from './model/ModelViewer';
 
+import TextField from './textfield/TextField';
```
gr812b, line 16: Remove testing code!

front-end/src/components/App.tsx
```diff
@@ -45,6 +47,8 @@ const App = () => {
     return () => clearTimeout(timer); 
   }, [successMessage]);
 
+  const [value, setValue] = useState('');
```
gr812b, line 50: ^^

front-end/src/components/textfield/TextField.tsx
```diff
@@ -0,0 +1,42 @@
+import styles from './TextField.module.scss';
+
+interface TextFieldProps {
+  title: string;
+  label?: string;
+  placeholder?: string;
+  defaultValue?: string;
+  value?: string;
+  onChange?: (value: string) => void;
```
gr812b, line 9: Renaming this to setValue would be more accurate. Also both value and setValue should be required fields.

front-end/src/components/textfield/TextField.tsx
```diff
@@ -0,0 +1,42 @@
+import styles from './TextField.module.scss';
+
+interface TextFieldProps {
+  title: string;
+  label?: string;
+  placeholder?: string;
+  defaultValue?: string;
```
gr812b, line 7: I think defaultValue would automatically be populated by passing in a value prop. so should be unrequired now

gr812b: Oh yeah not even used in your props, just remove here

front-end/src/styles/_variables.scss
```diff
@@ -20,6 +21,7 @@ $baja-colour: #75151E;
   --text-colour: #{$text-colour};
 
   --light-grey: #{$upload-stroke};
+  --alt-text: #{$alternate-text};
```
gr812b, line 24: I grouped these by "theme" colours and default colours. IMO alt-text could be dictated by theme, and so should be grouped with the first bunch, and then just add in one placeholder value (any colour you want) for the light theme variant

gr812b (changes requested):
Looks great, just a few minor details

front-end/src/components/textfield/TextField.tsx
camdnnn: updated

front-end/src/styles/_variables.scss
camdnnn: I don't really understand what you want

gr812b:
Can you also show variant without label?

camdnnn:
> Can you also show variant without label?

image

front-end/src/styles/_variables.scss
camdnnn: I think I fixed it

front-end/src/components/textfield/TextField.module.scss
```diff
@@ -0,0 +1,39 @@
+@use '@styles/_fonts';
+
+.textField {
+  @include fonts.text-medium;
+  display: flex;
+  flex-direction: column;
```
gr812b, line 6: Is this doing smt?

gr812b (changes requested):
Should have a default placeholder text such as 'Value' or smt.

Text in the input should be left alligned I feel. Just have some sort of padding on the left of it to offset by a certain amount. Figma is wrong in this case IMO. This is just for when its longer it looks a bit silly
image

When the parent element is small enough, the text field seems to not shrink down (some reason the minimum is 195) I did some investigating and have no fckin clue why this happens, but looks to me like width:100%; fixed it on the input class
image

gr812b (approved):

PR #208, opened 2025-01-21
Option square UI redesign

t-wing11:
Created the option square for the edit graph page.

Have a default image being put in right now
Also a default label "Option Title"

Passing in a clicked boolean for when this parent component is created.

Normal
image

Hover
image

Clicked
image

Clicked hovered
image


t-wing11:
Needs code cleanup but functionality is there

front-end/src/components/optionSquare/optionSquare.module.scss
```diff
@@ -0,0 +1,45 @@
...
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    border-radius: $border-radius;
+    border: 1px solid var(--background);
+    background-color: var(--tertiary);
+    color: var(--text-color);
+    width: 215px;
+    height: 180px;
```
gr812b, line 17: http://wiki.mcmasterbaja.ca/en/DAQ/Data-Viewer/CSS-Guidelines

Please use rem for this! :D

front-end/src/styles/_variables.scss
```diff
@@ -6,7 +6,7 @@ $background: #222222;
 $accent: #BB0808;
 $text-colour: #FFFFFF;
 $upload-stroke: #636363;
-
+$hover-colour: mix($tertiary, #FFFFFF, 90%);
```
gr812b, line 9: IMO this isn't really a theming thing and should instead be replicated with filter: brightness(110%). This avoids having too many theme colours. I've done it a bit for a few cases but I think we should avoid adding many more colours here if we can

front-end/src/components/optionSquare/optionSquare.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const OptionSquare: React.FC<OptionSquareProps> = ({ label = "Option Title", illustration = graphImage, isClicked}) => {
+    const [clicked, setClicked] = useState(isClicked);  // To track the click state
+
+    const handleClick = () => {
+        setClicked(prev => !prev); // Toggle clicked state
+    };
+    
+    return (
+<div className={`${styles.optionSquare} ${clicked ? styles.active : ''}`}  onClick={handleClick}>
```
gr812b, line 19: Use classNames (import as cx) for stuff like this pls!

front-end/src/components/optionSquare/optionSquare.tsx
```diff
@@ -0,0 +1,28 @@
+import { useState } from 'react';
+import styles from './optionSquare.module.scss';
+import graphImage from '../../assets/graph_image.png';
```
gr812b, line 3: Use alias

front-end/src/components/optionSquare/optionSquare.module.scss
```diff
@@ -0,0 +1,45 @@
...
+        img {
+            width: 100%;
+            height: 100%;
+            object-fit:fill;
+        }
+    }
+      .label {
+        text-align: center;
+        color: var(--text-colour);
+        padding-top: 5px;
```
gr812b, line 43: Use rem

front-end/src/components/optionSquare/optionSquare.module.scss
```diff
@@ -0,0 +1,45 @@
+@use '@styles/_fonts';
+@use '@styles/_variables';
+
+$border-radius: 8px;
```
gr812b, line 4: I made border-radius a var since i used it in other calcs, I think here its fine to leave as is? Just use 8px directly

front-end/src/components/optionSquare/optionSquare.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface OptionSquareProps {
+  label?: string;
+  illustration?: string;
+  isClicked?: boolean;
```
gr812b, line 8: I don't think any three of these should be nullable

front-end/src/components/optionSquare/optionSquare.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const OptionSquare: React.FC<OptionSquareProps> = ({ label = "Option Title", illustration = graphImage, isClicked}) => {
```
gr812b, line 11: React.FC is an outdated way of giving props typing. See other PRs

gr812b (changes requested):
I put a lot of comments but most are like 5sec fixes, just being nitpicky on the first ticket for standards :D

front-end/src/components/optionSquare/optionSquare.tsx
t-wing11: Valid

front-end/src/components/optionSquare/optionSquare.tsx
t-wing11: This was wrong actually, i removed this since we wont be using a default icon for these anyways

front-end/src/components/optionSquare/optionSquare.tsx
t-wing11: Yup i see, chatgpt led me astray

front-end/src/styles/_variables.scss
t-wing11: Didnt know u could do that, fixed

front-end/src/components/optionSquare/optionSquare.tsx
t-wing11: Feel like a junior dev learning from the senior rn, didnt know u could this either lmao

front-end/src/components/optionSquare/optionSquare.module.scss
```diff
@@ -0,0 +1,44 @@
...
+    justify-content: center;
+    align-items: center;
+    //use px for border
+    border-radius: 8px;
+    border: 1px solid var(--background);
+    background-color: var(--tertiary);
+    color: var(--text-color);
+    width: 13.435rem;
+    height: 11.25rem;
+
```
gr812b, line 17:
```suggestion
transition: filter 0.1s ease-in-out;
```
Would consider adding this to make the transition a bit less jarring

front-end/src/components/optionSquare/optionSquare.module.scss
```diff
@@ -0,0 +1,44 @@
...
+    &:hover {
+      filter: brightness(1.1);
+   }
+
+   &.clicked {
+    border: 3px solid var(--primary);
```
gr812b, line 23: I think @draeve might have some feedback for this, imo looks strange and not very clear it is selected.

front-end/src/components/optionSquare/optionSquare.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const OptionSquare = ({ label, illustration, clicked, setClicked}: OptionSquareProps) => {
+    
+    const handleClick = () => {
+        setClicked(!clicked);
+    };
+
+    return (
+        <div className={cx(styles.optionSquare, { [styles.clicked]: clicked })} onClick={handleClick}>
+        <div className={styles.illustration}>
+                {illustration ? <img src={illustration} alt="Illustration" /> : null}
```
gr812b, line 21: Now that illustration is a required field you should be able to get rid of this I think
```suggestion
                <img src={illustration} alt="Illustration" />
```

gr812b (commented):
Basically just pickiness on styling

front-end/src/components/optionSquare/optionSquare.tsx
```diff
@@ -0,0 +1,30 @@
+import { useState } from 'react';
+import styles from './optionSquare.module.scss';
+import cx from 'classnames';
+import checkMark from '../../assets/icons/checkmark.svg';
```
gr812b, line 4: use alias

front-end/src/assets/graph_image.png
gr812b, line 1: Remove

gr812b (changes requested):
fyucu

gr812b (approved):

PR #210, opened 2025-01-24
Add Sidebar

camdnnn:
Adds the sidebar component to the UI
image
image


front-end/src/components/sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const SidebarItem = ({ icon, text, onClick }: SidebarItemProps) => {
+        return (
+            <button className={styles.item} onClick={onClick}>
+                <div className={styles.iconContainer}>
+                    <img className={styles.icon} src={icon} />
+                </div>
+                <span className={styles.text}>{text}</span>
+            </button>
+        )
+    }
```
gr812b, line 31: When making a child component, I would have this be defined outside of the Sidebar component yet in the same file. Just separates them a bit to show SidebarItem is a standalone. Also seems to be convention

front-end/src/components/sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,63 @@
...
+            </div>
+            <div className={styles.body}>
+                <div className={styles.section}>
+                    <span className={styles.sectionTitle}>OVERVIEW</span>
+                </div>
+                <SidebarItem icon={homeIcon} text="Home" onClick={() => console.log('home')} />
+                <SidebarItem icon={bookmarkIcon} text="Bookmarked" onClick={() => console.log('bookmarked')}/>
+                <SidebarItem icon={uploadIcon} text="Upload Data" onClick={() => console.log('upload data')}/>
+                <SidebarItem icon={downloadIcon} text="Download Data" onClick={() => console.log('download data')}/>
+                <SidebarItem icon={folderIcon} text="File Browser" onClick={() => console.log('file browser')}/>
```
gr812b, line 50: I would look into using the <nav /> component here for standards / accessibility.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

gr812b (commented):
I just came to see how you were doing it

front-end/src/components/sidebar/Sidebar.tsx
camdnnn: Not all of them are navigation though, since some open modals or bookmark (idk what that would do)

front-end/src/components/sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,214 @@
...
+        cursor: pointer;
+    }
+
+    .toggleOpen {
+        @extend .toggle;
+        left: calc(100% - #{$toggleWidth});
+        opacity: 0;
+    }
+
+    .toggleClosed {
```
gr812b, line 208: Collapsed instead of closed

front-end/src/components/sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+// Sidebar item component
+const SidebarItem = ({ icon, text, onClick }: SidebarItemProps) => {
+    return (
+        <div className={styles.sidebarItem} onClick={onClick}>
+            <div className={styles.iconContainer}>
+                <img className={styles.icon} src={icon} alt="Icon" />
+            </div>
+            <span className={styles.text}>{text}</span>
+        </div>
```
gr812b, line 30: Make a button for easier

gr812b (changes requested):
image
Spacing looks a tad off under "Overview"
image

gr812b (changes requested):
Please make sure the /old still works, currently broken

front-end/src/components/sidebar/Sidebar.module.scss
camdnnn: naming was updated

front-end/src/components/sidebar/Sidebar.tsx
camdnnn: made into button

camdnnn:
> image Spacing looks a tad off under "Overview" image


image

gr812b (approved):

Issue #71, opened 2023-11-30
Fix Header Shifting


camdnnn:
Currently when you open folders in the file select menu, it causes the headers to shift positions. This should be changed so that the opening and closing of folders does not affect the positon of the headers

gr812b:
Would be nice to create our own library instead of rely on the outdated one for fileStorage in general. Currently this is what forces us to have to use `--force` when running `npm i`

Issue #139, opened 2024-11-15
Redo Frontend FileStorage


gr812b:
Currently our frontend fileStorage component depends on an outdated package, which is currently opening us up to some major vulnerabilities, causing usage of the `--force` tag when installing packages. The component is actually quite simple and we do not need to depend on this package.

This will include:
- Html table attributes
- Dynamic sizing of cells
- Carrying over javascript code for selecting files

PR #213, opened 2025-01-29
173 UI file table

gr812b:
File table component! One to one replacement for the FileStorage. Related to https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/139, which now only needs the removal of the old code.

Would be possible to fix https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/71 as well, if this is desired. The fix would make the date column a fixed width on the right, and then the size would follow next. This would replace the tables auto sizing which handles it based on content width.

Further, I will remove test code before merge in, this is for reviewers to test locally easily.

Normal
image

Expanded folders
image

Selected files
image


front-end/src/components/fileTable/FileTable.tsx
```diff
@@ -0,0 +1,250 @@
...
+    const assignComputedValues = (folder: Folder) => {
+        folder.children.forEach((child) => {
+            if ('children' in child) {
+                assignComputedValues(child);
+            }
+        });
+
+        if (folder.key) {
+            folder.size = sizeMap[folder.key] || 0;
+            folder.date = dateMap[folder.key]?.sort()[0] || ''; // Earliest date
```
hydrowoxy, line 84: maybe use 
```
const dates = dateMap[folder.key] || [];
folder.date = dates.length > 0 ? dates.toSorted()[0] : '';
```
 to avoid changing the original dates array

hydrowoxy (approved):
Very cool!! Not sure if this is too nitpicky or even really an issue but sometimes if you click something the bolding of the text shifts everything including column titles over a little like so
image
image
Looks great wowie!

front-end/src/components/fileTable/FileTable.tsx
gr812b: Unsure why I would need it in original order but I suppose good practice, implemented!

PR #214, opened 2025-01-29
182 UI Popout sidebar

gr812b:
Created the right sidebar component, which resizes the rest of the content of the page dynamically.

image


camdnnn (commented):
On my machine there is no resizer tab, also there is no way to reopen the window once it is closed. I'm not sure if these are the desired behaviours or not?

Screenshot with no tab to control window
image

Screenshot after hitting x and window closes with no way to reopen
image

camdnnn:
> On my machine there is no resizer tab, also there is no way to reopen the window once it is closed. I'm not sure if these are the desired behaviours or not?
> 
> Screenshot with no tab to control window image
> 
> Screenshot after hitting x and window closes with no way to reopen image

Update, I didn't have all the changes pulled, so side tab is included and works but still have no way of reopening sidebar once it has been closed

front-end/src/components/rightSidebar/RightSidebar.module.scss
```diff
@@ -0,0 +1,92 @@
...
+      position: absolute;
+      left: -0.5 * $resizerWidth;
+      top: 50%;
+      transform: translateY(-50%);
+      width: $resizerWidth;
+      height: 4rem;
+      border-radius: 20px;
+      background-color: var(--background-sec);
+      cursor: ew-resize;
+      position: relative;
```
hydrowoxy, line 69: Alr have position as absolute above

hydrowoxy (approved):
I have the same thing as cam with no option to re-open it once it's closed. Also, if I drag it to the left and let go, the animation still slowly chases my mouse for a while (see below)


https://github.com/user-attachments/assets/2daf44ae-db50-4091-ade9-0d7114563526

gr812b:
The sidebar isn't meant to re-open, it's a pop out! You click a button and it appears, open / closed state is handled by parent component

front-end/src/components/rightSidebar/RightSidebar.module.scss
gr812b: Good catch

gr812b:
> I have the same thing as cam with no option to re-open it once it's closed. Also, if I drag it to the left and let go, the animation still slowly chases my mouse for a while (see below)
> 
>  2025-02-23.19-38-01.2.mp4

Regarding the resizing going slowly, thats caused by the graph being funky on its resize observer, and unrelated to this. It will be fixed when we update the graph component.

Issue #176, opened 2025-01-10
UI - Graph Header


gr812b:
**Feature Description**
Create the graph header component. This includes a title and two buttons. Share dashboard opens up share modal. Configure can be left to do nothing for now. Also background gradient.

Props:
- Title

![image](https://github.com/user-attachments/assets/3f5af358-6988-4ef3-bd59-6d731f063441)

PR #215, opened 2025-02-13
UI - Graph Header (#176)

hydrowoxy:
Graph header component from the Figma with title prop, configure and share buttons, gradient background

image


gr812b (approved):
Just a few things to fix then good to go

PR #217, opened 2025-02-20
159 generate errorresponse via built in helper function to custom errors

hydrowoxy:
New BajaException superclass (extends RuntimeException) with a toErrorResponse() function that is passed down to all it's subclasses/all our custom errors. Simplified mappers to work with new structure.


backend/src/main/java/com/mcmasterbaja/exceptions/InvalidArgumentException.java
gr812b, line 1: Can you use the same pattern of defining local vars for the message and code as the others do?

gr812b (approved):
Wow look at all that code cleanup, how lovely is that!

PR #219, opened 2025-02-21
174 UI   download modal

gr812b:
Added the new download modal which looks super snazzy!

image
image


front-end/src/components/downloadModal/downloadModal.module.scss
```diff
@@ -0,0 +1,38 @@
+@use '@styles/_fonts';
+
+.downloadModal {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 0 2rem;
```
t-wing11, line 7: Is this a mistake `padding: 0 2rem;`?

front-end/src/components/downloadModal/downloadModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+const apiToFiles = (apiFiles: FileInformation[]): File[] => {
+  return apiFiles.map((file) => {
+    const keyParts = file.key.split('/');
+    const name = keyParts[keyParts.length - 1];
+    const extension = name.split('.')[name.split('.').length - 1];
+    return {
+      key: file.key,
+      name,
+      size: file.size,
+      date: '2021-01-01', // TODO: Make this dynamic
```
t-wing11, line 88: What goes into making this dynamic? Cause this would just mark every file you download be from this date and thats annoying, similar to the issue the daq box has

gr812b: https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/222

front-end/src/components/downloadModal/downloadModal.module.scss
gr812b: No! I just told you abt this in person 😘

camdnnn (commented):
Extremely nitpicky but the folder icon and text look a bit off center to me. I think that it is because the centering is done based on the full icon when I think that it would look better if done discounting the extra lip. 

What is currently looks like
image

Lines included to demonstrate what I mean. It is currently centered between the green and purple but I think it would look between centered using the red and green
image

hydrowoxy (approved):

gr812b:
> Extremely nitpicky but the folder icon and text look a bit off center to me. I think that it is because the centering is done based on the full icon when I think that it would look better if done discounting the extra lip.
> 
> What is currently looks like image
> 
> Lines included to demonstrate what I mean. It is currently centered between the green and purple but I think it would look between centered using the red and green image

There is a 0% chance I fix this... 😄

PR #220, opened 2025-02-22
175 UI   upload select files modal

gr812b:
Added visuals for the form along with basic functionality - This includes receiving the users inputs and returning them via a prop to whichever parent handles the information flow. Should be easy to link from there.

image


gr812b:
Should update this to only show user bins

front-end/src/components/presetFilesModal/PresetFilesModal.module.scss
```diff
@@ -0,0 +1,65 @@
+@use '@styles/_fonts';
```
t-wing11, line 1: nit: fix some spacing in this file

front-end/src/components/presetFilesModal/PresetFilesModal.module.scss
```diff
@@ -0,0 +1,65 @@
...
+        &::-webkit-scrollbar-track {
+          background: none;
+        }
+        &::-webkit-scrollbar-thumb {
+          background-color: var(--light-grey);
+          border-radius: 4px;
+          border: 1px solid var(--tertiary);
+        }
+        &::-webkit-scrollbar-thumb:hover {
+          background-color: #999; // TODO: Remove this colour
```
t-wing11, line 55: What does this mean, should you be removing this color now or is it later thing?

front-end/src/components/presetFilesModal/PresetFilesModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+      setExistingFiles(apiToFiles(files));
+    }
+
+    fetchFiles();
+  }, []);
+
+  const handleSubmit = async () => {
+    if (selectedFiles.length === 0 && uploadedFiles.length === 0) {
+      // TODO: Replace alert with popup or toast notification
+      alert('Please select or upload at least one file');
```
t-wing11, line 34: Is this still that default alert from before? Just wondering what it currently looks like

front-end/src/components/presetFilesModal/PresetFilesModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <h2 className={styles.title}>...Or select existing ones</h2>
+            <div className={styles.tableWrapper}>
+              <FileTable files={existingFiles} selectedFiles={selectedFiles} setSelectedFiles={setSelectedFiles} />
+            </div>
+          </div>
+        </div>
+
+        <Button 
+          onClick={handleSubmit}
+          textSize={'2rem'}
```
t-wing11, line 69: Shouldn't this be something handled in the style or are you overriding the textsize here?

t-wing11 (approved):
Just some comments and a couple questions. Otherwise looks good to me, you can address the nits if you want to

front-end/src/components/presetFilesModal/PresetFilesModal.tsx
gr812b: Just the built in alert that pops out from up top in your browser, we will replace this when we add a better alert system ourselves

front-end/src/components/presetFilesModal/PresetFilesModal.tsx
gr812b: Overriding default textSize with this

front-end/src/components/presetFilesModal/PresetFilesModal.module.scss
gr812b: I wasn't a fan of using non-built in colours so that theming changes automatically, but im not sure what to do with this one tbh

PR #221, opened 2025-02-22
180 UI - Accordion

hydrowoxy:
Accordion component which can collapse/expand to hide/show children, with arrow that rotates and line that raises and lowers (didn't really animate this...? not sure if this is the desired effect). Also not sure if it should be backgroundless, just assumed based on the figma.

image
image

I will remove the test code/return the app to normal before merging it!! Playwright is not happy because I nuked App.tsx!

### DIFF
--- a/front-end/src/components/accordion/Accordion.module.scss
+++ b/front-end/src/components/accordion/Accordion.module.scss
@@ -77,7 +77,6 @@
         transform: translateY(0.5rem);
     }
     100% {
-        opacity: 1;
         transform: translateY(0); 
     }
 }
--- a/front-end/src/components/accordion/Accordion.module.scss
+++ b/front-end/src/components/accordion/Accordion.module.scss
@@ -23,6 +23,7 @@
         width: 100%;
         height: 0.1rem;
         background-color: white;
+        animation: slideUpDown 0.5s ease-in-out forwards; 
     }
 
     .content {
@@ -34,6 +35,7 @@
             padding-bottom: 1rem;
             gap: 1rem;
             flex-wrap: wrap;
+            animation: slideContent 0.2s ease-out forwards; 
         }
 
         > * {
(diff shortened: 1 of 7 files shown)


front-end/src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <div className={styles.accordion}>
+      <button
+        className={styles.header}
+        onClick={() => setIsOpen((prev) => !prev)}
+      >
+        <img
+          src={icons.chevronDown}
+          alt="Toggle"
+          className={`${styles.arrow} ${isOpen ? styles.open : ""}`}
```
gr812b, line 21: Make sure to use classnames (imported as cx) instead for the naming of these. You can check out other components for reference

gr812b (changes requested):
If possible, please look into adding an animation! I think it'll go a long way for cleaning it up - other than that code looks great!

Edit: Also if possible indent the scss more! Take advantage of that to keep the structure clear in the scss as well

front-end/src/components/accordion/Accordion.module.scss
```diff
@@ -0,0 +1,63 @@
...
+        width: 100%;
+        height: 0.1rem;
+        background-color: white;
+    }
+
+    .content {
+        max-height: 0;
+        overflow: hidden;
+        padding-left: 1rem;
+        transition: max-height 0.3s ease-in-out, padding-bottom 0.3s ease-in-out;
```
gr812b, line 31: Can you make this 0.5s? Feels a bit nicer at that speed imo

front-end/src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+          className={cx(styles.arrow, { [styles.open]: isOpen })}
+        />
+        <span className={styles.title}>{title}</span>
+      </button>
+      <div
+        ref={contentRef}
+        className={cx(styles.content, { [styles.expanded]: isOpen })}
+        style={{
+          maxHeight: contentHeight,
+          transition: "max-height 0.3s ease-in-out",
```
gr812b, line 43: I think you can delete this line as you have it in scss

gr812b (approved):
Just remember to remove the edits to app and that slight animation change, looks super sick now I am very excited

PR #224, opened 2025-02-25
151 refactor try with resources

hydrowoxy:
Try-with-resources logic built in to getReader and getWriter in Analyzer; uses functional interfaces to accept a lambda expression and run it inside the try-with-resources block. getReader variant to work with multiple readers for InterpolatorPro. ~~"safe" readNext/readAll/flush to avoid unnecessary try-catches in the lambda expressions since they'd otherwise be out of scope for the interceptor to see IOExceptions.~~ Business logic refactored out of the main analyze functions so that they can be annotated with SneakyThrows and exceptions can be handled properly by the interceptor


gr812b (changes requested):
Let me start by saying the safety of the system is much improved in many key ways. This will definitely make the system more reliable for sure, which is fantastic!

That being said, I think the readability really suffers here - All of these methods having that extra indentation twice each time bogs it down quite a bit, and changing to safeReadNext feels unintuitive? 

I think this is a hard problem to solve, I spent some time thinking of solutions and the best I have is to separate out getting the readers and writers in a safe way. Something along the lines of this:

```java
private void processFile(String inputFile, String outputFile) {
        getReader(inputFile, reader -> {
            getWriter(outputFile, writer -> {
                try {
                    doStuff(reader, writer);
                } catch (IOException e) {
                    handleProcessingException(e); // Or just handle via interceptors
                }
            });
        });
    }
```
This at the very least keeps the business logic as its own thing in the doStuff() function, which is nice, but it might get out of hand with something like interpolaterProAnalyzer, hard to say.

I also have no clue how to exactly solve this safeReadNext to be in a way that doesn't feel odd, although potentially within the above refactor maybe you can put the interceptor on the doStuff() function and it would be happy.

Finally, I feel as though Analyzer.java is getting a bit out of hand itself as well. Realistically, this reader / writer stuff is not reliant on the analyzer stuff, so it might be a good idea to separate this out into its own file. Might want to like pseudo-code this out before fully commiting cause it might look stupid, but having so many things within the analyzer class is a tad odd (generally I like to keep it to 5 ish functions at most in one file)

backend/src/main/java/com/mcmasterbaja/analyzer/Analyzer.java
```diff
@@ -29,16 +36,60 @@ public void extractParams(AnalyzerParams params) {
...
+      for (String filePath : filePaths) {
+        try {
+          FileReader fileReader = new FileReader(filePath);
+          BufferedReader bufferedReader = new BufferedReader(fileReader);
+          CSVReader reader = new CSVReaderBuilder(bufferedReader).withSkipLines(0).build();
+          readersMap.put(filePath, reader);
+        } catch (IOException e) {
+          exceptions.add(
+              new InvalidInputFileException("Failed to read input file: " + filePath, e));
+        }
```
gr812b, line 63: I think this inner try catch is unnecesary, any failing of opening readers should definitely immediately throw an error. Then in the catch exception e we can just directly throw the error there

gr812b (changes requested):

gr812b (approved):

PR #225, opened 2025-02-26
Title Card 

t-wing11:
Adding title card which is comprised of:

Title with gradient on some of the words
Description

Graph which is using a GPS Speed vs time csv
- Added a small parser which parses and passes the data to an overriden chart options
- Overrides default chart options to make the graph look the right way for the title 
image

### DIFF
--- a/front-end/src/components/titleCard/titleCard.module.scss
+++ b/front-end/src/components/titleCard/titleCard.module.scss
@@ -1,0 +1,51 @@
+@use '@styles/_fonts';
+@use '@styles/_variables';
+
+.titleCard {
+    display: flex;
+    justify-content: space-between;
+    gap: 7rem;
+    align-items: center;
+    width: 85.94rem;
+    height: 21.69rem;
+    background-color: var(--tertiary);
+
+    .textContainer {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        justify-content: center;
+        opacity: 100%;
+        width: 27.83rem;
+        height: 12.63rem;
+        border: 1px solid var(--background);
+        margin-left: 3rem;
+
+        .title {
+            @include fonts.title-huge;
+            text-align: center;
+            width: 90%;
+            color: white;
+            border: 1px solid var(--background);
+        }
+
+        .description {
+            @include fonts.text-large;
+            text-align: center;
+            width: 80%;
+            color: white;
+            border: 1px solid var(--background);
+            margin-top: 10px;
+        }
+    }
+
+    .graph {
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        width: 40.19rem;
+        height: 16.47rem;
+        background-color: var(--secondary);
+        margin-right: 5rem;
+    }
+}
--- a/front-end/src/components/titleCard/titleCard.module.scss
+++ b/front-end/src/components/titleCard/titleCard.module.scss
@@ -18,7 +18,6 @@
         opacity: 100%;
         width: 27.83rem;
         height: 12.63rem;
-        border: 1px solid var(--background);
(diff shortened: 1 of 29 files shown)


front-end/src/styles/_fonts.scss
```diff
@@ -75,7 +75,7 @@ $base-font-size: 16px;
 
 @mixin title-huge {
   font-size: $base-font-size * 4;
-  line-height: $base-font-size * 6.25;
+  line-height: 1;
```
gr812b, line 78: This def should not be changed I don't think, what's the reasoning?

gr812b (commented):
I'll let you clean up the code first before going in-depth, but looks mostly good to me

front-end/src/styles/_fonts.scss
t-wing11: The line height was really large, had to lower it to align with the figma

front-end/src/components/titleCard/titleCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+                data: chartData
+            }
+        ]
+    };
+    return (
+        <div className={styles.titleCard}>
+           <div className={styles.textContainer}>
+            <div className={styles.title}>
+            <span className={styles.highlight1}>VISUALIZE</span> YOUR <br/>
+            DATA, <span className={styles.highlight2}>YOUR WAY </span>
```
gr812b, line 73: Highlight should be on way only, not your

front-end/src/styles/_fonts.scss
gr812b: 1 is def not right, should be based on base font size. Fine to lower but make sure its a real number otherwise will lead to issues

gr812b (commented):
Just some cleanup, I'll look into adding multiple series and different data

front-end/src/components/titleCard/titleCard.module.scss
```diff
@@ -0,0 +1,63 @@
...
+        display: flex;
+        justify-content: center;
+        background: 'transparent';
+        align-items: center;
+        width: 40.19rem;
+        height: 16.47rem;
+        margin-right: 5rem;
+        border-bottom: 1px solid grey;
+        overflow: hidden;
+        resize: both;
```
gr812b, line 61: Remove this resize stuff

front-end/src/components/titleCard/titleCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+    useEffect(() => {
+        setChartData(parseData(titleCardData));
+    }, []);
+
+    const options: Highcharts.Options = {
+        ...defaultChartOptions, 
+        chart: {
+            ...defaultChartOptions.chart,
+            backgroundColor: 'transparent',
+            height: 250
```
gr812b, line 40: Coding this in feels unnecessary?

gr812b:
image
Updated graph visuals

front-end/src/components/titleCard/titleCard.tsx
gr812b: I fixed it such that the chart container is just based on the size of this, so at least it only needs to change in one spot rather than two

front-end/src/components/titleCard/titleCard.module.scss
gr812b: Oops I think I accidentally commited this

front-end/src/components/titleCard/titleCard.tsx
t-wing11: oh mb, i had this correct then stashed the changes and forgot to do it

gr812b (approved):

PR #231, opened 2025-03-16
178 UI   graph preset page

gr812b:
Still need to fix the resizing issue, other than that should be ok!


front-end/src/components/sidebar/Sidebar.tsx
```diff
@@ -51,10 +59,13 @@ const Sidebar = () => {
...
+                {/* <SidebarItem icon={bookmarkIcon} text="Bookmarked" onClick={() => console.log('bookmarked')}/> */}
                 <SidebarItem icon={uploadIcon} text="Upload Data" onClick={() => openModal('upload')}/>
                 <SidebarItem icon={downloadIcon} text="Download Data" onClick={() => openModal('download')}/>
-                <SidebarItem icon={folderIcon} text="File Browser" onClick={() => console.log('file browser')}/>
+                {/* <SidebarItem icon={folderIcon} text="File Browser" onClick={() => console.log('file browser')}/> */}
+                <SidebarItem icon={cubeIcon} text="Model Viewer" onClick={() => onIconClick('IMU')}/>
+                <SidebarItem icon={newGraphIcon} text="Legacy Create Graph" onClick={() => setModal('Create')}/>
+                <SidebarItem icon={historyIcon} text="Legacy UI" onClick={() => onIconClick('old')}/>
             </div>
             <div className={styles.footerHeader}>
```
camdnnn, line 70: I think it would be better if possible to use the already used openModal function for the create button instead of adding in a prop

front-end/src/components/sidebar/Sidebar.tsx
```diff
@@ -51,10 +59,13 @@ const Sidebar = () => {
             </div>
             <div className={styles.bodyItems}>
                 <SidebarItem icon={homeIcon} text="Home" onClick={() => onIconClick('')}/>
-                <SidebarItem icon={bookmarkIcon} text="Bookmarked" onClick={() => console.log('bookmarked')}/>
+                {/* <SidebarItem icon={bookmarkIcon} text="Bookmarked" onClick={() => console.log('bookmarked')}/> */}
                 <SidebarItem icon={uploadIcon} text="Upload Data" onClick={() => openModal('upload')}/>
                 <SidebarItem icon={downloadIcon} text="Download Data" onClick={() => openModal('download')}/>
-                <SidebarItem icon={folderIcon} text="File Browser" onClick={() => console.log('file browser')}/>
+                {/* <SidebarItem icon={folderIcon} text="File Browser" onClick={() => console.log('file browser')}/> */}
+                <SidebarItem icon={cubeIcon} text="Model Viewer" onClick={() => onIconClick('IMU')}/>
```
camdnnn, line 66: I feel like either IMU shouldn't be here or Map should be added. Both are currently accessed through routing. Though I think it should be neither and they should be both part of the multiple graphs

camdnnn (changes requested):
Seems pretty good to me, just have some minor comments for the sidebar

front-end/src/components/sidebar/Sidebar.tsx
gr812b: Agreed but they aren't added through graphs so will add map here

front-end/src/components/sidebar/Sidebar.tsx
gr812b: Reason behind doing this is that its the old modal, which won't be sticking around and so this will be removed. It would be a larger pain in the ass to rework it rn just to rework the actual functionality again later

camdnnn (approved):

PR #239, opened 2025-05-23
Added Outlier Ranges

camdnnn:
Updated the existing Delete Outliers and Strict Timestamp analyzers to use ranges instead of singular values


gr812b (changes requested):
Strict timestamp prob shouldn't need a min or max, keep to remove outliers and then just call both in some combined

gr812b (approved):

PR #249, opened 2025-07-12
Graph Styling

gr812b:
Added graph styling to better match figma

image

### DIFF
--- a/front-end/src/components/legacy/views/Chart/Chart.module.scss
+++ b/front-end/src/components/legacy/views/Chart/Chart.module.scss
@@ -4,10 +4,6 @@
     resize: both;
     overflow: hidden;
 
-    > * {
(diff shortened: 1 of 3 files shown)


front-end/src/components/legacy/views/Chart/Chart.module.scss
```diff
@@ -26,9 +22,8 @@
     }
 
     .chart {
-        
         :global(.highcharts-root) {
-            font-family: "HelveticaNeueW01-77BdCnObl" !important;
+            font-family: 'Roboto', sans-serif !important;
```
hydrowoxy, line 26: Oh my goodness

hydrowoxy:
red

hydrowoxy (approved):
Oh my god I found the button
Good job it's so figma pilled now

PR #261, opened 2025-08-12
59 edit graph

gr812b:
image

Fixed multiple series and added edit bar styling


hydrowoxy (commented):
Yay

hydrowoxy (approved):
Yay (with approval)

PR #270, opened 2025-08-31
266 Remove legacy code

gr812b:
Removed a ton of legacy code. Please refer to the old version of code if you want to see how it was done previously.


hydrowoxy (approved):
so much cleanup!

PR #272, opened 2025-09-02
Fix verbose error messages

hydrowoxy:
AnalyzerExceptionInterceptor catches and re-throws custom errors to avoid long nested  messages

Before Example
image

After Example
image

So now only the message from the first place where it failed will remain


gr812b (commented):
Oops I should probably comment not approve

e2e/testers/upload.ts
hydrowoxy, line 1: idek

gr812b:
> image better!

I would try to include the details field as well since it gives a bit more context here that's useful

hydrowoxy:
image

Retains more detail. technical stuff visually muted so it doesn't look like a lot of text

hydrowoxy:
image

Better!

gr812b (approved):
Wow much good the final product is mint

PR #274, opened 2025-09-03
273 fix e2e testing

hydrowoxy:
Basic fixes to e2e testing for the new UI

Some things could be improved to be more robust like properly detecting when a graph was created vs just looking for the dashboard title, and maybe testing more presets? instead of just the speed one


gr812b (commented):
You'll need to update WEB_SERVER_URL from the \old endpoint to get it to work.

PR #278, opened 2025-09-10
269 bring back multiple graphs

hydrowoxy:


hydrowoxy:
Made a separate ticket for fixing presets because it is taking me too long and I want to get this merged in

PR #280, opened 2025-09-16
Fixed subtitle to show -- need to fix CSS/ styling to look better

switzmab:
- Subtitle can now be edited and properly displayed
- Styling for 'subtitle' will need to be fixed next


hydrowoxy (approved):
Nice looks like it works! The styling is actually fine this way as this was the intended appearance

PR #282, opened 2025-09-17
improved styling a little bit

hans-hc:
fixed up some styling (border bevel, font/color, background color when drop down), otherwise didnt mess with it too much

### DIFF
--- a/front-end/src/components/ui/accordion/Accordion.module.scss
+++ b/front-end/src/components/ui/accordion/Accordion.module.scss
@@ -13,6 +13,8 @@
 
 .accordion {
     position: relative;
+    background-color: var(--background);
+    border-radius: 0.5rem;
 
     &::before {
         content: '';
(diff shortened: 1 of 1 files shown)


gr812b (commented):
http://wiki.mcmasterbaja.ca/en/DAQ/Data-Viewer/CSS-Guidelines

Wanted to put this here, from this specifically I would check the colour stuff - it helps centralize these instead of having them all spread out. Helps a lot with themes!

Also, helps to throw an image in your PR desc when wokring on a front end thing, just to show the final, pretty result. Good work!

hans-hc:
got it, thanks! I'll tweak the code a little more to use the centralized color stuff. Should I be making another PR after I do that?

JAWalmsley:
Looks good to me.
While we're in the accordion styling though, the title and body should probably share a border, not have separate "bubbles" with rounded corners between them
image
I think it looks better more like this (quick photoshop)
image
Might require putting the border styling on the accordion container instead of the button itself

JAWalmsley (approved):
Gorgeous

PR #305, opened 2025-10-26
IMU Acceleration Vectors

camdnnn:
Adding the ability to display acceleration vectors in the IMU player

image

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -81,13 +81,7 @@
         <dependency>
             <groupId>com.fazecast</groupId>
             <artifactId>jSerialComm</artifactId>
-            <version>[2.0.0,3.0.0)</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.fazecast</groupId>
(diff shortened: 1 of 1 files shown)


front-end/src/lib/modelUtils.ts
```diff
@@ -14,45 +14,138 @@ const parseCSV = (data: string) => {
...
+  return new ArrowHelper(dir.clone().normalize(), origin, length, color);
+};
+
+const getScaledLength = (value: number, maxAccel: number, maxLength: number) => {
+  return (Math.abs(value) / maxAccel) * maxLength;
+};
+
+const updateArrow = (arrow: ArrowHelper, vec: Vector3, length: number, radius: number) => {
+  const dir = vec.clone().normalize();
+  const origin = dir.clone().multiplyScalar(radius);
```
gr812b: Wouldn't origin always be in center?

gr812b: Ooh I see, makes sense!

PR #308, opened 2025-10-30
Fixed long title causing icon problems

hans-hc:
- clamp title to 2 lines (min-width: 0 so it shrinks nicely) <--chatgpt helped with the webkit stuff
- keep fileCount from wrapping/shrinking by pinning it to the top
- should be aligned with single-line cards so rows stay level
- not a function change but polished up tiny spacing

### DIFF
--- a/front-end/src/components/composite/presetList/presetCard/PresetCard.module.scss
+++ b/front-end/src/components/composite/presetList/presetCard/PresetCard.module.scss
@@ -51,8 +51,7 @@
 
     .title {
         flex: 1 1 auto;
-        min-width: 0;
-        line-height: 1.25;         
+        min-width: 0;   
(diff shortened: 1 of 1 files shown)


front-end/src/components/composite/presetList/presetCard/PresetCard.module.scss
```diff
@@ -42,17 +42,36 @@
...
+        gap: 0.75rem;
         width: 100%;
         padding-bottom: 0.6rem;
         color: var(--text-colour);
     }
 
+    .title {
+        flex: 1 1 auto;
+        min-width: 0;
+        line-height: 1.25;
```
gr812b: Line height should be used in he fonts only probably

PR #309, opened 2025-11-01
IMU Playbar

camdnnn:
Added playbar to IMU

Also cleaned up the  modelUtils functions by moving many into the class which greatly reduced the number of parameters

image

Ignore the actual timestamp values as the data I am using has malformed timestamps

### DIFF
--- a/front-end/src/components/ui/playbar/Playbar.tsx
+++ b/front-end/src/components/ui/playbar/Playbar.tsx
@@ -41,7 +41,7 @@ export const Playbar = ({ replayController, times }: PlaybarProps) => {
         setCurrentIndex(event.currentIndex);
       } else if (event.type === ReplayEventType.Finished) {
         setIsPlaying(false);
-        setCurrentIndex(replayController.data.length - 1);
+        setCurrentIndex(times.length - 1);
       }
     });
     return cleanup;


front-end/src/components/legacy/model/ModelViewer.tsx
```diff
@@ -3,95 +3,69 @@ import { Canvas } from '@react-three/fiber';
...
+import { fetchData, ModelReplayController, ReplayModelSubscriber } from '@lib/modelUtils';
 import './modelViewer.css';
 import { ApiUtil } from '@lib/apiUtils';
-import { ReplayEvent, ReplayEventType, StateType, replayData } from '@types';
+import {ReplayEventType, replayData } from '@types';
 import { showInfoToast } from '@components/ui/toastNotification/ToastNotification';
+import { Playbar } from '@components/ui/playbar/Playbar';
 
 const ModelViewer = () => {
   const objRef = useRef<THREE.Group>(undefined);
```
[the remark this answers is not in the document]
camdnnn: No, this breaks stuff

front-end/src/lib/modelUtils.ts
```diff
@@ -1,18 +1,29 @@
...
+  StateType
+} from '@types';
+import {
+  Quaternion,
+  ArrowHelper,
+  Vector3,
+  Box3,
+  Sphere,
+  Group,
+  Object3D
```
gr812b: Remove

front-end/src/lib/modelUtils.ts
```diff
@@ -22,302 +33,308 @@ const combineData = (
...
+    this.listeners.clear();
+  }
+}
 
-    if (this.scene) {
-      Object.values(this.accelVectors).forEach(vec => {
-        this.scene?.remove(vec);
-        vec.dispose();
-      });
+export class ReplayModelSubscriber {
```
gr812b: This is nice, clean way to handle lots of guys at once. Maybe consider moving into another file than the replayController or smt, lots in one file. Not pressing though

front-end/src/components/ui/playbar/Playbar.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  padding: 0.75rem 1.5rem;
+  width: 100%;
+
+  .playPause {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 2.5rem;
+    height: 2.5rem;
+    padding-left: 3px; /* to center the play icon visually */
```
gr812b: I know I added this but I think it messes up the pause button visually

Issue #104, opened 2024-04-01
Folder upload / download


gr812b:
Would be related to https://github.com/McMaster-Baja-Racing/Better-Data-Viewer/issues/98 where you can select a whole folder to upload, which would accept any of the valid file types and store it in that folder on the backend. Further, you should be able to select whole folders at a time on fileSelect for download

PR #315, opened 2025-11-30
104 folder upload download

CowMan999:
Solves #104 by allowing folder uploads via drag&drop and the file <input> prompt


front-end/src/components/composite/uploadModal/uploadModal.tsx
```diff
@@ -16,6 +16,7 @@ interface UploadModalProps {
 export const UploadModal = ({ isOpen, onClose }: UploadModalProps) => {
   const [files, setFiles] = useState<File[]>([]);
   const { refetch } = useFiles();
+  const [allowFolder, setAllowFolder] = useState(false);
```
gr812b: Is this toggle needed here? I like the fact you can disable it, for the presetFilesModal, but for upload I think it should be okay to always allow folders as well as normal files too

PR #331, opened 2026-01-14
Fixed Upload Data Modal and Download Data Modal

### LINKED ISSUE #281: Make the UI more responsive to different screen sizes
Doesn't need to work on mobile devices or anything crazy, but could be a bit better when the window gets squished or whatnot. Right now some of the styling gets funky with certain components.

hans-hc:
as title indicates, fixed the sizing glitches when the window is squished. Also slightly adjusted the download data modal so by default full window the box isnt surpassing the height of the window.
image
image

### DIFF
--- a/front-end/src/components/composite/downloadModal/downloadModal.module.scss
+++ b/front-end/src/components/composite/downloadModal/downloadModal.module.scss
@@ -14,11 +14,60 @@
   }
 
   .tableWrapper {
-    height: 35rem;
+    height: 30rem;
     @include mixins.scrollbar-dark;
   }
 
   .submitButton {
     margin-top: 1.25rem;
+  }
+}
+
+/* Responsive styles for smaller desktop windows */
+@media screen and (max-width: 1400px) {
+  .downloadModal {
+    width: 60rem;
+
+    .tableWrapper {
+      height: 32rem;
+    }
+  }
+}
+
+@media screen and (max-width: 1200px) {
+  .downloadModal {
+    width: 50rem;
+    padding: 0 1.5rem;
+
+    .tableWrapper {
+      height: 30rem;
+    }
+
+    .submitButton {
+      margin-top: 1rem;
+    }
+  }
+}
+
+@media screen and (max-width: 1000px) {
+  .downloadModal {
+    width: 40rem;
+    padding: 0 1rem;
+
+    .tableWrapper {
+      height: 25rem;
+    }
+  }
+}
+
+@media screen and (max-width: 850px) {
+  .downloadModal {
+    width: 90vw;
+    max-width: 35rem;
+    padding: 0 1rem;
+
+    .tableWrapper {
+      height: 20rem;
+    }
   }
 }
--- a/front-end/src/components/composite/downloadModal/downloadModal.module.scss
+++ b/front-end/src/components/composite/downloadModal/downloadModal.module.scss
@@ -17,11 +17,6 @@
     height: 30rem;
     @include mixins.scrollbar-dark;
   }
-
-  .submitButton {
-    margin-top: 1.25rem;
-    margin-bottom: 0.5rem;
-  }
 }
 
 /* Responsive styles for smaller desktop windows */
@@ -30,12 +25,12 @@
     width: 60rem;
 
     .tableWrapper {
-      height: 32rem;
+      height: 30rem;
     }
   }
 }
 
-@media screen and (max-width: 1200px) {
+@media screen and (max-width: 1024px) {
   .downloadModal {
     width: 50rem;
     padding: 0 1.5rem;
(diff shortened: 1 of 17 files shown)


hydrowoxy (approved):
You've overlooked so much many things fool!

### Extremely **infinite** _major_ functional hugepick is the spacing under the upload/download button when I have it in split screen (see below)

I can't believe you would be so careless in your blumbering.... big buffoon I swear. ***Don't** let it happen **again***.

Image

Image

Both are half my screen but the download one is without a doubt a **_TON_** more cramped below the red button, and the upload one has wayyy too much padding below it? Bakaaaa!!!

front-end/src/components/composite/uploadModal/uploadModal.module.scss
```diff
@@ -15,5 +15,33 @@
...
+    margin-bottom: 0.5rem;
+  }
+}
+
+@media screen and (max-width: 1200px) {
+  .uploadModal {
+    width: 35rem;
+    padding: 0 1.5rem;
+
+    .submitButton {
```
gr812b: Can you put this styling directly into our button component? Might make it cascade to the whole codebase a bit better

front-end/src/components/composite/downloadModal/downloadModal.module.scss
```diff
@@ -14,11 +14,61 @@
...
+    width: 60rem;
+
+    .tableWrapper {
+      height: 32rem;
+    }
+  }
+}
+
+@media screen and (max-width: 1200px) {
+  .downloadModal {
```
gr812b: This sorta stuff could also be done in baseModal maybe so itll affect this, the upload and the preset and every other modal down the line

gr812b (changes requested):
You blithering fool! You absolutely definitively must implement this in the UI folder of components and do as much as you can in there lest it cascade in a catastrophically faulty manner. More work for you <3
Looks terrible as well. This is the antithesis of the correct end product. You have created something so diametrically opposite to what we expected that you absolutely need to start worrying. 
Be concerned about mobile sizes whether you want to or not. It will be seamless to use on mobile without a ground up redesign.

front-end/src/components/simple/howItWorks/HowItWorks.module.scss
```diff
@@ -114,21 +114,117 @@
...
+      gap: 2rem;
+
+      .stepContent {
+        max-width: 400px;
+      }
+
+      .stepImage {
+        max-width: 400px;
+        min-height: 250px;
+        height: 250px;
```
gr812b: no px!

front-end/src/components/simple/howItWorks/HowItWorks.module.scss
```diff
@@ -114,21 +114,117 @@
...
+        .stepText {
+          font-size: 0.9rem;
+        }
+      }
 
-      .stepContent, .stepImage {
+      .stepImage {
         max-width: 100%;
+        min-height: 200px;
+        height: 200px;
```
gr812b: No px!

front-end/src/components/ui/uploadForm/UploadForm.module.scss
```diff
@@ -15,6 +15,23 @@
...
+    min-width: 30rem;
+    min-height: 18rem;
+  }
+
+  @media screen and (max-width: 1000px) {
+    min-width: 25rem;
+    min-height: 16rem;
+  }
+
+  @media screen and (max-width: 850px) {
```
gr812b: 850 corresponds to what device?

gr812b (changes requested):
The 1000 and 850 breakpoints feel a bit odd, last I believe there are standard ones like 1400+ for ultrawide, down to 1024 for normal, then down to 768 for handhelds and <768 for rest. Might wanna just change those. Might also be good to write a mixin in the @styles folder to centralize these values

hydrowoxy (changes requested):
what kai said

front-end/src/components/simple/footer/Footer.module.scss
```diff
@@ -7,45 +7,58 @@
     background: var(--background);
```
hydrowoxy: Footer looks a bit cramped on the right, Even if i elongate the window this persists

Image

front-end/src/components/composite/presetFilesModal/PresetFilesModal.tsx
```diff
@@ -37,47 +35,26 @@ export const PresetFilesModal = ({ onClose, isOpen, onSubmit, currentSources }:
...
-                setSelectedFiles={setSelectedFiles} 
-              />
-            </div>
-          </div>
+        <h2 className={styles.title}>Select a file</h2>
+        <div className={styles.tableWrapper}>
+          <FileTable 
+            files={getFolders(files || [])} 
+            selectedFiles={selectedFiles} 
+            setSelectedFiles={setSelectedFiles}
```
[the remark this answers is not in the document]
gr812b: No, only on mobile sizes. Neccesary on desktop

gr812b (commented):
Just gonna review functionality first

Image
No space between button and upload box. Also did you make the default one full screen smaller? Should remain at original size for full sized screens.

At 1000px and 1024px it changes size. Why at both so close? Just stick to one (1024)

Image
This one is also smaller. Keep the original size for full-sized screens. Also button spacing wrong here too.

Image
When shrinking the screen it can look like this, need reactivity for the table

Image
For the select modal, the table shrinks properly. Strange that its diff here and on download. Also as stated elsewhere need that functionality

gr812b (changes requested):
Not gonna go through it all but you should use your defined variables and mixin at each of these spots

PR #341, opened 2026-01-14
removed accel curve from front end and backend enum

hydrowoxy:


front-end/src/types/AnalyzerTypes.ts
gr812b: I concur remove pls

backend/src/main/java/com/mcmasterbaja/model/AnalyzerType.java
gr812b: Remove this too? He's dead is he not

